PR #5, opened 2022-10-21
fix!: Switch `Parser` outputs to Associated Types

epage:
General wisdom is that generics are used for input types and associated types are used for outputs.

In our case, we are trying to make the compiler happy for Geal/nom#1535


epage:
Note: this was implemented and closed as part of figuring out how to workaround an issue I was having with constraining generic parameters in #1535 which I found another way to do (and then rejected the entire approach).

This doens't mean this idea doesn't have merit, its just the driving factor for trying it is gone.

PR #47, opened 2022-12-19
feat(parser): impl Parsers for literals

epage:
We now support
- `&str` and `&[u8]` as `tag` parsers
- `char` as a `char` parser and `u8` as a `one_of` parser
- u8 and char ranges as `one_of` parsers

This is to allow reducing noise and making more natural to write a parser.  The downside is when running into type inference issues.

The other downside is that this prevents us from moving `Parser` and `IResult` into a `-core` crate.  However, I would expected we'd need most of `nom::input` in that as well which would be harder to keep compatibility on I suspect.

Fixes Geal/nom#1417
Fixes Geal/nom#1408

### DIFF
--- a/examples/iterator.rs
+++ b/examples/iterator.rs
@@ -64,10 +64,7 @@ fn main() {
   // - we can get the remaining input afterwards, like with the `try_fold` trick
   let mut nom_it = iterator(
     data,
-    terminated(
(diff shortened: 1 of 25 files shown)


epage:
A problem I ran into with this is that I was defining symbols left and right and didn't pay attention to when they were contained in a tuple which was breaking my parsing because it was then treated as a sequence, rather than a compound `one_of`.

PR #173, opened 2023-02-17
fix(error)!: Deprecate `dbg_dmp` in favor of `trace`

epage:


epage:
Not sure why clippy was fine before but not now but an upcoming PR will take care of it

Issue #203, opened 2023-03-04
`Partial::complete` does not actually set its state as complete


lo48576:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### rust version

rustc 1.67.1 (d5a82bbd2 2023-02-07)

### winnow version

master

### Minimal reproducible code

```rust
fn main() {
    let mut i = Partial::new(&b""[..]);
    assert!(Partial::<&[u8]>::is_partial_supported());

    assert!(i.is_partial(), "incomplete by default");
    let incomplete_state = i.complete();

    // This assertion fails with `master` branch, but I think it should succeed.
    assert!(!i.is_partial(), "the stream should be marked as complete");
}
```

### Steps to reproduce the bug with the above code

1. `cargo run`.

### Actual Behaviour

The last assertion (`!i.is_partial()`) fails.

### Expected Behaviour

All assertions (including the last `!i.is_partial()`) succeeds.

### Additional Context

[The document for `StreamIsPartial::complete()`](https://docs.rs/winnow/0.3.3/winnow/stream/trait.StreamIsPartial.html#tymethod.complete) says "Mark the stream is complete", so I expect `StreamIsPartial::is_partial()` would be false after calling that but actually it is not.
If this behavior is intentional, please make the document more clear about what `complete()` is expected to do.

---

I tried to write a patch <https://github.com/winnow-rs/winnow/compare/main...lo48576:winnow:partial-complete> (#204), but it makes `length_value_test()` at `src/multi/tests.rs` and doc test of `length_value()` at `src/multi/mod.rs` fail. All other tests passed.

The below is run at the commit 8d6465d7c1f47708af076226f769c1f11dd791f7.
```
$ cargo test
(snip)
failures:

---- multi::tests::length_value_test stdout ----
thread 'multi::tests::length_value_test' panicked at 'assertion failed: `(left == right)`
  left: `Err(Backtrack(Error { input: Partial { input: [], partial: false }, kind: Eof }))`,
 right: `Err(Backtrack(Error { input: Partial { input: [], partial: true }, kind: Complete }))`', src/multi/tests.rs:506:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


[...]

PR #204, opened 2023-03-04
Set `Partial` complete after `Partial::complete()` is called

lo48576:
The patch and test fixes for #203.


epage:
... can't believe that fell through the cracks.

Thanks for fixing this!

epage:
v0.3.4 is now released

PR #205, opened 2023-03-05
fix: Make things compilable without std nor alloc

lo48576:
## Description
The crate (`master` branch) does not compile with nostd.

## Steps to reproduce
1. Run `cargo check --no-default-features` and/or `cargo test --no-default-features`.
2. Run `cargo check --no-default-features --features=alloc` and/or `cargo test --no-default-features --features=alloc`.

## Expected result
Both `check` and `test` pass even when `std` feature is disabled.

## Actual result
`check` and `test` fail when `std` feature is disabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ required-features = ["alloc"]
 [[example]]
 name = "css"
 test = true
+required-features = ["alloc"]
 
 [[example]]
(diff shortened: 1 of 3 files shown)


epage:
Thanks!  The CI checks for this were brittle which I have now made more resilient so this shouldn't happen in the future.

Issue #223, opened 2023-03-24
performance regression in winnow 0.4.0


martinohmann:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### rust version

rustc 1.68.0 (2c8cc3432 2023-03-06)

### winnow version

v0.4.0

### Minimal reproducible code

`Cargo.toml`:

```toml
[package]
name = "winnow-regression"
version = "0.1.0"
edition = "2021"

[[bench]]
name = "bench"
harness = false

[dev-dependencies]
criterion = "0.4.0"
winnow = "0.3.6" # Change to 0.4.0 to see the regression in the benchmark
```

`benches/bench.rs`:

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion, Throughput};
use std::time::Duration;
use winnow::{character::multispace0, error::Error, Parser};

fn bench(c: &mut Criterion) {
    let mut group = c.benchmark_group("parse");

    let input = "\t ".repeat(1024 * 1024);

    group.throughput(Throughput::Bytes(input.len() as u64));
    group.warm_up_time(Duration::from_secs(5));
    group.measurement_time(Duration::from_secs(10));

    group.bench_function("multispace0", |b| {
        multispace0::<&str, Error<&str>>.parse_next(&input).unwrap();
        b.iter(|| black_box(multispace0::<&str, Error<&str>>.parse_next(&input).unwrap()))
    });

    group.finish();
}

criterion_group!(benches, bench);
criterion_main!(benches);
```

### Steps to reproduce the bug with the above code

Run `cargo criterion` using `winnow` 0.3.6 and 0.4.0 to see the performance difference.

### Actual Behaviour

Benchmark result with winnow 0.4.0:

```text
parse/multispace0       time:   [2.4784 ms 2.4805 ms 2.4831 ms]
                        thrpt:  [805.45 MiB/s 806.29 MiB/s 806.97 MiB/s]
                 change:
                        time:   [+45.529% +45.820% +46.096%] (p = 0.00 < 0.05)
                        thrpt:  [-31.552% -31.422% -31.285%]
                        Performance has regressed.
```

### Expected Behaviour

Benchmark result with winnow 0.3.6:

```
[...]

PR #227, opened 2023-03-24
docs: Add `hcl-edit` to the list of parsers using `winnow`

martinohmann:
Our recent discussion in https://github.com/winnow-rs/winnow/issues/223 pushed me to finally make the initial release of `hcl-edit` which I've been holding of for too long already since I got carried away with missing details.

The crate is far from complete, but the parser implementation is functional and actually quite fast.

So I figured I'll add it to the list of parsers for others to explore real-world uses of winnow.

I hope this does not scare away people from winnow since HCL is quite a big and messy piece to parse, given the three sub-languages it has and some other quirks like the absense of keywords (for example, `true` can be an identifier or a boolean, depending on the context) :wink:.

### DIFF
--- a/src/_topic/mod.rs
+++ b/src/_topic/mod.rs
@@ -19,6 +19,7 @@
 //! See also parsers written with `winnow`:
 //!
 //! - [`toml_edit`](https://crates.io/crates/toml_edit)
+//! - [`hcl-edit`](https://crates.io/crates/hcl-edit)
 
 pub mod arithmetic;
 pub mod error;


epage:
> I hope this does not scare away people from winnow since HCL is quite a big and messy piece to parse, given the three sub-languages it has and some other quirks like the absense of keywords (for example, true can be an identifier or a boolean, depending on the context) 😉.


Wow, that makes me really interested in reading your thoughts on and challenges in using winnow to implement hcl-edit

Issue #231, opened 2023-04-14
Mechanisms to manipulate parser state in a `map`-like fashion


martinohmann:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### winnow version

v0.4.1

### Describe your use case

See the discussion thread https://github.com/winnow-rs/winnow/discussions/230#discussioncomment-5620269

The general idea is to get access to the state of a `winnow::stream::Stateful` input in a `map`-like function so that it can be updated while mapping a parser's output.

Right now, only simple use cases are supported, like the one shown in the [`Stateful` type docs](https://docs.rs/winnow/latest/winnow/stream/struct.Stateful.html).

### Describe the solution you'd like

A method on the `Parser` trait that allows to modify the state in a `map`-like operation.

Ideally something like `map_with_state` that would both provide the output of the previous parser, and a reference to the `state` inside the `Stateful` so that it could be updated.

There was another idea in https://github.com/winnow-rs/winnow/discussions/230#discussioncomment-5620269 to generalize this to not only support the `Stateful` input type, but any input type by having something like `map_with_input` to expose a reference to the whole input, not just the `state` in a `Stateful`.

martinohmann:
@epage feel free to rename the issue to something better, I was struggling to find a good title.

martinohmann:
I was thinking about it a bit more and checked the existing API of the `Parser` trait. We could design this in a way that better integrates with what already exist. Taking `recognize`/`with_recognized` and `span`/`with_span` as examples, we could:

- add a `stream` method which yields the a `Stream` containing the input that produced the output of the previous parser
- add `with_stream`, which yields a tuple of parser output and `Stream`

Then you could simply access this via a `map` call. This might be enough to both support the `Stateful` use case and the use case mentioned in https://github.com/winnow-rs/winnow/discussions/230#discussioncomment-5620026.

I can try to compile a draft PR for this approach in the next few days and play around with it a bit to see if this would actually work in the way I envision.

martinohmann:
I'm also not too nailed down on the naming yet. Maybe `output_stream`/`with_output_stream` would be a better fit to make it more obvious what's contained in the returned `Stream`.

[...]

PR #233, opened 2023-04-17
feat: Add `output_stream`/`with_output_stream` to `Parser`

martinohmann:
This is a draft implementation to solve https://github.com/winnow-rs/winnow/issues/231

As noted in the initial issue, I'm not too settled on the naming here, there are other alternatives like `recognize_stream`/`with_recognized_stream`.

If you think the idea is worth pursuing further, I can also add some tests/examples for use cases like the following:

- use `Parser::output_stream().and_then(..)` to map another parser over the `Stream`.
- use `Parser::with_output_stream().map(...)` to access/update the `state` in a `Stateful` in the map closure.

Happy to hear your thoughts on this.

### DIFF
--- a/src/combinator/mod.rs
+++ b/src/combinator/mod.rs
@@ -159,7 +159,7 @@
 use crate::error::{ContextError, ErrMode, ErrorKind, FromExternalError, Needed, ParseError};
 use crate::lib::std::borrow::Borrow;
 use crate::lib::std::ops::Range;
-use crate::stream::{Location, Stream};
+use crate::stream::{Location, Stream, UpdateSlice};
 use crate::stream::{Offset, StreamIsPartial};
 use crate::trace::trace;
 use crate::trace::trace_result;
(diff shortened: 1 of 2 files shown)


martinohmann:
@epage any opinion about this?

src/parser.rs
```diff
@@ -258,6 +261,99 @@ pub trait Parser<I, O, E> {
...
+    /// assert_eq!(
+    ///     parser.parse_next(BStr::new("abcd,efgh")),
+    ///     Ok((BStr::new(""), BStr::new("abcd,efgh"))),
+    /// );
+    /// assert_eq!(
+    ///     parser.parse_next(BStr::new("abcd;")),
+    ///     Err(ErrMode::Backtrack(Error::new(BStr::new(";"), ErrorKind::Verify))),
+    /// );
+    /// ```
+    fn output_stream(self) -> OutputStream<Self, I, O, E>
```
epage, line 288: I think I prefer `recognize`, rather than `output` as that does clarify what the stream's range  is

src/combinator/mod.rs
```diff
@@ -972,6 +972,96 @@ where
...
+    I: UpdateSlice,
+{
+    fn parse_next(&mut self, input: I) -> IResult<I, I, E> {
+        match self.parser.parse_next(input.clone()) {
+            Ok((remaining, _)) => {
+                let offset = input.offset_to(&remaining);
+                let (remaining, slice) = input.next_slice(offset);
+                Ok((remaining, input.update_slice(slice)))
+            }
+            Err(e) => Err(e),
```
epage, line 1015: Should the streams be marked complete?

src/parser.rs
```diff
@@ -212,29 +216,29 @@ pub trait Parser<I, O, E> {
 
     /// Produce the consumed input with the output
     ///
-    /// Functions similarly to [recognize][Parser::recognize] except it
-    /// returns the parser output as well.
+    /// Functions similarly to [`recognize`][Parser::recognize] except it returns the parser output
```
epage, line 219: Unrelated changes will need to be reverted

src/parser.rs
epage, line 288: Could `length_value` be updated to use this?

src/parser.rs
```diff
@@ -258,6 +261,99 @@ pub trait Parser<I, O, E> {
...
+    /// assert_eq!(
+    ///     output_stream_parser.parse_next(BStr::new("1234")),
+    ///     consumed_parser.parse_next(BStr::new("1234")),
+    /// );
+    /// assert_eq!(
+    ///     output_stream_parser.parse_next(BStr::new("abcd")),
+    ///     consumed_parser.parse_next(BStr::new("abcd")),
+    /// );
+    /// ```
+    fn with_output_stream(self) -> WithOutputStream<Self, I, O, E>
```
epage, line 349: Have you considered trying these out on an extension trait in your crate first to explore their usage?

I am considering doing 0.5 in the next couple months, so we do have room to change things soon if we aren't thrilled with it.  I'm just exploring (and reminding myself) what the options are for these

epage (commented):
Be sure to read my comment about extension traits before applying other feedback

src/parser.rs
martinohmann: Yep, `recognize_stream` also feels more fitting to me. Will rename it when I polish the PR.

src/combinator/mod.rs
martinohmann: Maybe I'm missing something, but how would I mark a stream as complete?

src/parser.rs
martinohmann: Can do that, but I thought it would make sense to add the missing backticks around `recognize` as part of this PR to align it with the docs of the new methods since they are related.

src/parser.rs
martinohmann: I didn't consider that yet, no. But it's a great idea and it'll do this to see if that additional methods are really that useful as I imagine. Will report back once I did that.

I already played around with `output_stream` in my `hcl-edit` parser and it actually helped to simplify one of my use cases quite a bit. Didn't check how (if) `with_output_stream` would help simplifying the parsers that receive a `state` parameter yet, but I did a small PoC in a bigger unit test that I didn't bundle with the PR yet to verify that state updates within a `map` operation actually work the way I expect.

src/parser.rs
martinohmann: I'll have a look.

src/combinator/mod.rs
epage: See https://github.com/winnow-rs/winnow/blob/main/src/binary/mod.rs#L2487

src/parser.rs
epage: If so, that should be a separate commit

martinohmann:
@epage I think the new `&mut Input` implementation could change things here. I'm going to revisit this on top of the v0.5 release and see what we get.

martinohmann:
Closing as this PR got way out of sync with the current state of winnow. I don't have a compelling use case that justifies porting this to current winnow implementation. Probably not needed anyways after `Input` was made `&mut` a while back.

Issue #237, opened 2023-04-27
Large tuples are a performance footgun


epage:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### rust version

1.68

### winnow version

0.4.0

### Minimal reproducible code

```rust
(a, b).map(...).parse_next(input)
```

### Steps to reproduce the bug with the above code

TBD

### Actual Behaviour

Slow

### Expected Behaviour

Fast

### Additional Context

See #230

Options
- Reduce the number of tuple elements supported to not encourage bad behavior
- The struct macro from #82

epage:
> Hi there @epage!
> 
> In #223 I promised you some feedback about using `winnow` to create the parser for [`hcl-edit`](https://docs.rs/hcl-edit). I didn't get to it in the past few weeks due to other obligations, but now I've got a little bit of time to note down some of the things I learnt.
> ## Ergonomics
> 
> First of all I'd like to point out that the ergonomics of using `winnow` are much better than what I observed when initially starting out with `nom`. I like it a lot that `&str`, `char`, `byte` etc. also implement the `Parser` trait, which avoids a lot of wrapping in other combinators like `char()` or `tag()`.
> 
> Also, converting some of the combinators into methods on the `Parser` trait really helps ergnonomics.
> 
> ```rust
> parser.context("foo").context("bar).context("baz").parse(input)
> ```
> 
> is certainly easier to follow than
> 
> ```rust
> context("foo", context("bar", context("baz", parser)(input)
> ```
> 
> ## Type inference
> 
> Type inference works really well in most cases, but the `many*` and `separated*` combinators in particular require to help the compiler via turbofish here and there. This is sometimes a bit cumbersome and adds noise since these functions have a ton of type parameters, but only the parameter with the `Accumulate` trait bound needs a type hint.
> 
> For example, I'm parsing whitespace interleaved with comments as follows, which wouldn't typecheck without the turbofish:
> 
> ```rust
> pub(super) fn ws(input: Input) -> IResult<Input, ()> {
>     (multispace0, many0::<_, _, (), _, _>((comment, multispace0)))
>         .void()
>         .parse_next(input)
> }
> ```
> 
[...]

PR #238, opened 2023-04-27
feat(seq): 'seq' macro for easier skipping

### LINKED ISSUE #82: An easier way to discard fields in a tuple

### winnow version

0.2.0

### Describe your use case

Some sequences of parsers (via tuples) become long and it can be a mess to deal with the resulting values as some fields are there for parsing purposes and don't matter.  Judicious use of `preceded` and `terminated` help but that also makes it harder to read.

Ideally, I could also say to `cut` in the middle without creating a new tuple.

A macro for populating a struct
- https://docs.rs/combine/latest/combine/macro.struct_parser.html
- https://crates.io/crates/nom-derive
- https://crates.io/crates/nom-fields

_No response_

### Additional Context

_No response_

epage:
Fixes #82

### DIFF
--- a/examples/css/parser.rs
+++ b/examples/css/parser.rs
@@ -1,3 +1,4 @@
+use winnow::combinator::seq;
 use winnow::prelude::*;
 use winnow::token::take_while;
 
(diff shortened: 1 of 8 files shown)


src/macros.rs
```diff
@@ -76,3 +76,111 @@ macro_rules! succ (
...
+///     )),
+/// );
+/// ```
+#[macro_export]
+#[doc(alias = "tuple")]
+#[doc(alias = "preceded")]
+#[doc(alias = "terminated")]
+#[doc(alias = "pair")]
+#[doc(alias = "struct_parser")]
+macro_rules! seq {
```
epage, line 131: This needs to be added to the combinator table

epage:
So I got `seq!(Struct {})` working.  The problem is with tuples.  The only way I can find to make them work is if we do `().map` which means we hit the performance problem of #237 which I was trying to avoid

Issue #98, opened 2023-01-30
Consolidate parser ranged variants


epage:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### winnow version

0.2.0

### Describe your use case

Nom has `many1`, `many0`, and `many_m_n` functions.  Similar with other parts of the API.

### Describe the solution you'd like

What if instead we had a `IntoRange` trait that took in the different range types and single numbers.

Example:
```rust
many(tag("abc"), 0..) -> many0(tag("abc")) 
many(tag("abc"), 1..) -> many1(tag("abc")) 
many(tag("abc"), 1) -> many_m_n(tag("abc"), 1, 1) 
```

epage:
nom merged this as rust-bakery/nom#1608

Differences we want from that
- We want this for every applicable parser
- We want to deprecate the other versions

The main questions remaining
-  if we should keep `0` and `1` variants for convenience?
- Are there cases where we only want a `0`/`1` or a ranged?

epage:
Postponed as I want to remove the deprecated functions first

epage:
Potential parsers:
- many0 / many1
- fold_many0
- take, take_while0, take_while1, take_while_m_m
- take_till0., take_till1
- take_until0, take_until1
- many_till0, many_till1
- separated0

For `winnow::character` parsers like `digit0` / `digit1`, I'd like to find a different way of solving those.

epage:
I'm running into the dreaded
```
error[E0207]: the type parameter `I` is not constrained by the impl trait, self type, or predicates
  --> src/combinator/multi.rs:39:9
   |
39 | impl<P, I, O, E> crate::lib::std::ops::Mul<P> for Repeat
   |         ^ unconstrained type parameter

```
when using `Mul`

I was already thinking of having a fallback function.  Looks like it'll be required

epage:
Hmm, I feel like using functions for this, making the API fluent, is one part verbosity and one part being clever.  Maybe I should just fallback to doing this the original way

epage:
We now have
- `take_while`
- `repeat`
- `fold_repeat`

Still remaining
- `take_till0`., take_till1`
- `take_until0`, take_until1`
[...]

PR #247, opened 2023-05-02
fix(multi): Replace repeat0/repeat1 with repeat

epage:
This is a follow up to #241 for merging in the 0/1 variants

This is a step towards #98


epage:
@martinohmann  I'm looking for other opinions on whether it makes sense to replace `repeat0(p()` with `repeat(0.., p)` and `repeat1(p)` with `repeat(1.., p)` or if that is losing too much convenience for consistency / small API surface.

Still need to benchmark but I'm assuming the `repeat` wrapper will be inlined away to nothing.

martinohmann:
As long as performance does not suffer, I'm totally fine with only having the more flexible `repeat`. The new API surface isn't too bad.

PR #268, opened 2023-06-30
perf(parse)!: Switch from pure parsers to stream mutation

### LINKED ISSUE #72: Parsing takes 30% longer just by adding `Located` to the input
When evaluating location / span solutions and checking there performance, I found:
- `Located<I>` takes about a 30% performance hit over `I` without actually capturing spans
- [pori](https://crates.io/crates/pori)'s `Located` took the json benchmark from 1.1126 µs to 1.4623 µs
- [nom_locate](https://crates.io/crates/nom_locate)'s `Located` took the json benchmark from 1.4362 µs (pori) to 1.9520 µs
- `Stateful<I, ()>` has negligible performance impact
  - `Stateful` with `usize`, `(usize, usize)`, and `(usize, usize, usize)` mirrors the Located<I> performance and then continued to get slower from there.

See also
- #61
- #62
- #63
- #64

epage:
Everything is there for still writing pure parsers when people want to.

Remaining work
- [ ] Finish updating doc comments
- [ ] Finish removing `unpeek`
- [ ] Reproduce slowdown from #72 and see if its addressed

Fixes #72


examples/json/parser.rs
```diff
@@ -30,41 +29,41 @@ pub type Stream<'i> = &'i str;
...
+        string.map(JsonValue::Str),
         float.map(JsonValue::Num),
-        unpeek(array).map(JsonValue::Array),
-        unpeek(object).map(JsonValue::Object),
+        array.map(JsonValue::Array),
+        object.map(JsonValue::Object),
     ))
-    .parse_peek(input)
+    .parse_next(input)
 }
```
epage: This is an example of what a lot of the changes look like

(`unpeek` and `parse_peek` are approaches for making the old way work with the new API)

src/combinator/core.rs
```diff
@@ -116,19 +113,13 @@ where
...
-        }),
-    )
+    trace("cond", move |input: &mut I| {
+        if b {
+            f.parse_next(input).map(Some)
+        } else {
+            Ok(None)
+        }
+    })
 }
```
epage: In some ways, this makes things a lot cleaner

src/_topic/language.rs
```diff
@@ -301,15 +301,15 @@
 //!       '.',
 //!       opt(decimal)
 //!     ).recognize()
-//!   )).parse_peek(input)
+//!   )).parse_next(input)
 //! }
 //!
-//! fn decimal(input: &str) -> IResult<&str, &str> {
+//! fn decimal<'s>(input: &mut &'s str) -> PResult<&'s str> {
```
epage, line 307: This change can sometimes force explicit lifetimes

src/ascii/tests.rs
```diff
@@ -241,7 +241,7 @@ mod complete {
         let f = "βèƒôřè\rÂßÇáƒƭèř";
         assert_eq!(
             not_line_ending.parse_peek(f),
-            Err(ErrMode::Backtrack(Error::new(f, ErrorKind::Tag)))
+            Err(ErrMode::Backtrack(Error::new(&f[12..], ErrorKind::Tag)))
```
epage: This is an example of how error reporting changes with this.  We now report errors (and in generally leave the shared mutable stream) at the most specific location where the error occurred

src/combinator/branch.rs
```diff
@@ -145,26 +146,28 @@ macro_rules! alt_trait_impl(
...
-  ($it:tt, $self:expr, $input:expr, $err:expr, $head:ident $($id:ident)+) => (
-    match $self.$it.parse_peek($input.clone()) {
+  ($it:tt, $self:expr, $input:expr, $start:ident, $err:expr, $head:ident $($id:ident)+) => ({
+    $input.reset($start.clone());
+    match $self.$it.parse_next($input) {
       Err(ErrMode::Backtrack(e)) => {
         let err = $err.or(e);
-        succ!($it, alt_trait_inner!($self, $input, err, $($id)+))
+        succ!($it, alt_trait_inner!($self, $input, $start, err, $($id)+))
       }
```
epage, line 155: One of the main changes is if a user needs to backtrack, instead of preserving the original input, they take a `checkpoint` and `reset` the stream to it.

src/combinator/multi.rs
```diff
@@ -819,23 +838,25 @@ where
...
                 res = g(res, o);
-                input = i;
             }
             Err(ErrMode::Backtrack(_)) => {
-                return Ok((input, res));
+                input.reset(start);
+                return Ok(res);
             }
             Err(e) => {
                 return Err(e);
```
epage: Similarly, for repeating parsers, we have to `checkpoint` before each parse attempt and then `reset` when its the end of the error.

In general, this type of scheme means only a fraction of the code needs to do checkpointing.  The alternative would be every parser would need to perform a "transaction" and that seemed too invasive

epage:
@martinohmann I know you have limited time but I would appreciate feedback on this API change.  The primary motivation is to reduce overhead when using `Located` but I'm still working on verifying that it helps with that case.  At minimum, it doesn't seem to be hurting performance.  If it doesn't pan out, I'm assuming having a pure-functional API would be preferred and I'd reject this PR.

martinohmann:
> @martinohmann I know you have limited time but I would appreciate feedback on this API change. The primary motivation is to reduce overhead when using `Located` but I'm still working on verifying that it helps with that case. At minimum, it doesn't seem to be hurting performance. If it doesn't pan out, I'm assuming having a pure-functional API would be preferred and I'd reject this PR.

I had an hour of spare time and compiled an `hcl-edit` branch which uses the changes in this PR to see how this goes. I have to say that the migration was fairly easy and I was able to do the majority of necessary adjustments via search&replace (mostly changing `Input<'a>` to `&mut Input<'a>` and replacing `IResult` to `PResult`). I didn't look if there are new opportunities for performance optimizations or code simplifications yet, but just tried to make it work.

The branch is here: https://github.com/martinohmann/hcl-rs/compare/main...hcl-edit/winnow-stream-mutation

Benchmark results also look nice so far. I'm seeing ~15% performance improvements with this.

Current `main`:

```
parse/hcl-edit/deeply_nested.tf
                        time:   [18.422 µs 18.459 µs 18.511 µs]
                        thrpt:  [38.176 MiB/s 38.283 MiB/s 38.360 MiB/s]
parse/hcl-edit/large.tf time:   [1.8049 ms 1.8123 ms 1.8230 ms]
                        thrpt:  [44.719 MiB/s 44.983 MiB/s 45.166 MiB/s]
parse/hcl-edit/medium.tf
                        time:   [414.88 µs 415.94 µs 417.18 µs]
                        thrpt:  [34.418 MiB/s 34.521 MiB/s 34.609 MiB/s]
parse/hcl-edit/small.tf time:   [24.656 µs 24.675 µs 24.698 µs]
                        thrpt:  [38.342 MiB/s 38.379 MiB/s 38.409 MiB/s]
```

The `hcl-edit/winnow-stream-mutation` branch:

```
parse/hcl-edit/deeply_nested.tf
                        time:   [15.302 µs 15.306 µs 15.312 µs]
                        thrpt:  [46.153 MiB/s 46.169 MiB/s 46.182 MiB/s]
                 change:
                        time:   [-17.253% -16.961% -16.692%] (p = 0.00 < 0.05)
                        thrpt:  [+20.037% +20.425% +20.850%]
                        Performance has improved.
parse/hcl-edit/large.tf time:   [1.5211 ms 1.5299 ms 1.5418 ms]
                        thrpt:  [52.876 MiB/s 53.287 MiB/s 53.594 MiB/s]
                 change:
                        time:   [-16.262% -15.584% -14.790%] (p = 0.00 < 0.05)
                        thrpt:  [+17.357% +18.461% +19.420%]
                        Performance has improved.
parse/hcl-edit/medium.tf
                        time:   [355.48 µs 356.44 µs 357.55 µs]
                        thrpt:  [40.158 MiB/s 40.283 MiB/s 40.392 MiB/s]
                 change:
                        time:   [-14.652% -14.305% -13.994%] (p = 0.00 < 0.05)
                        thrpt:  [+16.270% +16.693% +17.168%]
                        Performance has improved.
parse/hcl-edit/small.tf time:   [20.405 µs 20.481 µs 20.590 µs]
                        thrpt:  [45.992 MiB/s 46.239 MiB/s 46.410 MiB/s]
                 change:
                        time:   [-17.433% -16.925% -16.355%] (p = 0.00 < 0.05)
                        thrpt:  [+19.552% +20.373% +21.113%]
                        Performance has improved.
```

epage:
Thanks for looking into that!

I might build off of that to see if there is further room for improvement.  If this is about return types, your error type is likely the remaining limiting factor.  We could instead store a `Stream::Checkpoint` inside of it and store the `cause` in `Context` which should shrink `ParseError` from being 9-10 pointers wide to 5 pointers wide.  Doubt its needed but storing content in a `Option<Box<Vec<Context>>>` might save another pointer-width without much or any performance impact.  Likely even the `Checkpoint` could be dropped because the `Stream` is left pointing to where the error is when everything is done.

The standard approach of just boxing all of the error wouldn't work well because that would make backtracking more expensive.

src/combinator/multi.rs
martinohmann: I think checkpoint/reset are better than the status quo since with that it's more explicit when and where input rewinding happens. See the changes to the `expr_inner` function in `parser/expr.rs`. I think with the checkpointing it's clearer what's going on now.

martinohmann:
Yeah, that's a good idea. I'll try that out on my branch the next time I have a spare hour and report back.

On another note: the excessive explict lifetime annotations I added are just a result of how I defined my own `PResult` for now because I was being lazy. They are not really a side effect of your changes (except in the cases that you pointed out further up).

epage:
I think the `cause` change slowed things down: https://github.com/epage/hcl-rs/tree/hcl-edt/winnow-error-cause

I think the `input` change was neutral on performance: https://github.com/epage/hcl-rs/tree/hcl-edt/winnow-error-input

epage:
As for reproducing the performance gains seen, I ran
```console
$ cargo bench -p benchmarks --bench parse -- parse/hcl-edit/large.tf
```
- Before: results ranged from 1.18ms - 1.30ms
- After: results ranged from 1.04ms - 1.14ms

(yeah, my machiner is jittery and I do about 4-5 runs of criterion)

src/ascii/tests.rs
martinohmann: I wonder if we could a wrapper that does this. I don't get where the offset of 12 comes from here and probably neither of the other users would.

epage: I'd be curious what you are thinking for what this wrapper would be.

In this case, the offset of 112 is the `\r` character which is what caused the error.

src/combinator/branch.rs
martinohmann: I find the checkpoint/reset concept very elegant. Maybe checkpoint need a different name. I instantly thought of postgres WAL checkpoints. Not too far off though. Maybe `freeze` could be a possible alternativ?

src/ascii/tests.rs
martinohmann: Oh, I see it now, nevermind then. I should go to bed...

src/combinator/branch.rs
epage: Yeah, I didn't want to block the experiment by waiting to come up with a name, so I just took what [combine named things](https://docs.rs/combine/latest/combine/stream/trait.ResetStream.html#associatedtype.Checkpoint).

I had also been considering `snapsnhot` though `save` and `bookmark` could also work.  `state` would overlap with `Stateful`.

For me, `freeze` carries the wrong annotation, that you are locking things down like in Python's `frozenmap`.

epage: Also, I wasn't sure if I could make checkpoints as small as they are.  I was surprised with what you can do with traits, like
- 8b3a1bcbab7cbe54be575b543e0fc783cf109e90 (passing an associated type as generic parameter to a trait)
- ade7912fec04e2a9b28d3900e9adc767313d9e96 (passing an associated type of the trait currently being defined to a generic parameter of a super trait)

The initialization-order aspects of those I find wild.

martinohmann: I learnt something new today: didn't know that you could use `as` in generic parameter position. 😃

martinohmann:
Another thing that might be useful to simplify the migration: I had to replace `Offset::offset_to` with `Offset::offset_from` in some places. Maybe don't remove `offset_to` but rather provide a default impl which uses `offset_from` under the hood.

src/combinator/branch.rs
epage: `winnow` has pushed my knowledge of what is possible...

epage:
> Another thing that might be useful to simplify the migration: I had to replace uses of `Offset::offset_to` with `Offset::offset_from` in some places. Maybe don't remove `offset_to` but rather provide a default impl which uses `offset_from` under the hood.

Good call.  I'll be evaluating the migration more when we are getting ready for release.  In that case, I might make changes to the previous release to help.

epage:
I've since split this out into individual PRs

I also just release 0.4.8 which includes `Parser::parse_peek`, `Stream::peek_token`, `Stream::peek_slice`, and `Offset::offset_from` which wrap existing functionality but make it easier to prep for the upgrade within the 0.4 branch

martinohmann:
This is great! Looking forward to v0.5. I'm also curious to see if it enables further performance improvements than what we already saw. Also, maybe the workaround described in https://docs.rs/winnow/latest/winnow/_topic/performance/index.html#built-time-performance to reduce build times in some cases isn't really needed anymore given that the overall signature of `FnMut` parsers is smaller now, which might simplify the generated types. At least for me, that would allow removing a substantial amount of boilerplate from my parser code. But maybe i'm completely off here.

epage:
>  overall signature of FnMut parsers is smaller now, 

I'm assuming its not enough of a difference in the signature but we'll see.

However, if you use `trace` to help debugging then you are always doing it anyways...

PR #301, opened 2023-08-05
Update dependencies to their latest version

dnaka91:
Noticed through `cargo-deny` that I have several duplicate dependencies. So I went ahead and updated
all to their latest version. Most notably is the updat of `anstyle` to the 1.0 release and replacing `anstyle-stream` with `anstream` (which seems to be a rename of the crate).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,21 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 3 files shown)


epage:
Thanks!

PR #317, opened 2023-08-17
perf(token): Speed up `take_until` under simd

epage:
This reverts commit 489aa2ace0ce77790cf8d2aa3fc0fcf8a8f7ccb2.

PR #86 "FindSubstring<&[u8]>: make use of `memchr::memmem`"

When testing under gitoxide, it turned out this slowed down gitoxide-specific micro-benchmarks from ~170ns to ~240ns


epage:
Important to note that for gitoxide, `take_until` looks for
- space
- newline
- `:`
- `>`
- ` <`
- PGP signature begin
- PGP signature end

So its a mix of single-character (the `memchr` shortcut) and actual substring search

PR #320, opened 2023-08-19
improve tutorial inner navigation links

joseluis:
Hi there! I figured out I could improve a little the navigation links system, and also use the preferred term [*Table of Contents*](https://english.stackexchange.com/questions/207086/table-of-contents-vs-table-of-content). I hope you like it. In any case please feel free to ask for any changes, modify it or close it. Thank you!

### DIFF
--- a/src/_tutorial/mod.rs
+++ b/src/_tutorial/mod.rs
@@ -1,6 +1,6 @@
 //! # Tutorial
 //!
-//! Table of Content
+//! Table of Contents
 
 pub mod chapter_0;
 pub mod chapter_1;
@@ -10,3 +10,17 @@ pub mod chapter_4;
 pub mod chapter_5;
 pub mod chapter_6;
 pub mod chapter_7;
+
+// Macro to generate the inner tutorial navigation links,
+// with optional previous and next links.
+macro_rules! tutorial_links {
+    (
+        $( previous: $prev_mod:ident $(,)? )?
+        $( next: $next_mod:ident $(,)? )?
+    ) => {
+        $( pub use super::$prev_mod as previous; )?
+        $( pub use super::$next_mod as next; )?
+        pub use crate::_tutorial as table_of_contents;
+    }
+}
+pub(self) use tutorial_links;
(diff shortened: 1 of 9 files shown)


src/_tutorial/mod.rs
```diff
@@ -10,3 +10,17 @@ pub mod chapter_4;
...
+macro_rules! tutorial_links {
+    (
+        $( previous: $prev_mod:ident $(,)? )?
+        $( next: $next_mod:ident $(,)? )?
+    ) => {
+        $( pub use super::$prev_mod as previous; )?
+        $( pub use super::$next_mod as next; )?
+        pub use crate::_tutorial as table_of_contents;
+    }
+}
```
epage, line 25: Not too much of a fan of this macro

src/_tutorial/mod.rs
```diff
@@ -10,3 +10,17 @@ pub mod chapter_4;
...
+
+// Macro to generate the inner tutorial navigation links,
+// with optional previous and next links.
+macro_rules! tutorial_links {
+    (
+        $( previous: $prev_mod:ident $(,)? )?
+        $( next: $next_mod:ident $(,)? )?
+    ) => {
+        $( pub use super::$prev_mod as previous; )?
+        $( pub use super::$next_mod as next; )?
```
epage, line 22: Is this actually meant to change anything?

The description doesn't make that clear and this commit mixes refactors (macro) with functional changes (pluralizing of "content" at minimum) which obscures whether other functional changes happened.

joseluis: The macro just helps to avoid boilerplate for generating the navigation links. The output is the same as before. The only other change was the pluralization of the TOC module which is done inside the macro. It could have been done in separate commits but the macro one would've overriden that anyway.

src/_tutorial/mod.rs
joseluis: Is there something in particular about the macro that doesn't convince you, or just the whole concept of creating a macro for generalizing the navigation boilerplate?

epage: I have a fairly high bar for the use of macros.  In my mind, this doesn't macro doesn't pay for itself in how much it makes development easier.

joseluis: Ok that's perfectly fine. We should close the PR then.

PR #332, opened 2023-09-15
Note that winnow doesn't provide byte-to-line-number conversions

9999years:
Follow-up from https://github.com/winnow-rs/winnow/discussions/330 to document the fact that `winnow` considers converting from span byte offsets to line numbers out of scope.

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -90,6 +90,13 @@ impl BStr {
 ///
 /// Spans are tracked as a [`Range<usize>`] of byte offsets.
 ///
+/// Converting byte offsets to line or column numbers is left up to the user, as computing column
+/// numbers requires domain knowledge (are columns byte-based, codepoint-based, or grapheme-based?)
+/// and O(n) iteration over the input to determine codepoint and line boundaries.
+///
+/// [The `line-span` crate](https://docs.rs/line-span/latest/line_span/) can help with converting
+/// byte offsets to line numbers.
+///
 /// See [`Parser::span`][crate::Parser::span] and [`Parser::with_span`][crate::Parser::with_span] for more details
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -288,6 +288,14 @@ pub trait Parser<I, O, E> {
 
     /// Produce the location of the consumed input as produced value.
     ///
+    /// Note that locations are returned as `usize` byte offsets, as converting these to line or
+    /// column numbers requires domain knowledge (are columns byte-based, codepoint-based, or
+    /// grapheme-based?) and O(n) iteration over the input (to determine codepoint and line
+    /// boundaries).
```
epage, line 294: Maybe extend #331 with some of this information

src/parser.rs
```diff
@@ -288,6 +288,14 @@ pub trait Parser<I, O, E> {
...
+    ///
+    /// Consider using [the `line-span` crate](https://docs.rs/line-span/latest/line_span/)
+    /// to convert from byte offsets to line and column numbers.
```
epage, line 297: I don't think I saw column handling in it which made me hesitant in recommending it.

9999years: That's correct, but I'm not aware of another crate that does handle column numbers, so I figure if someone wants those then `line-span` is as good a place to start as any. Would you prefer to omit this?

epage: Maybe just clarify that it helps to identify line numbers and column byte offsets (requiring users to convert to display column)

src/parser.rs
epage: Sorry, I was expecting this to be removed.  I'd rather have this all in one place rather than `Located`, `Parser::span`, and `Parser::with_span`

9999years: Oh, all on the `Located` type? OK.

epage:
Thanks!

PR #343, opened 2023-10-06
docs: Be explicit on offset vs index

### LINKED ISSUE #341: Out-of-bounds offset when input isn't long enough for parser

### rust version

rustc 1.72.1 (d5c2e9c34 2023-09-13)

### winnow version

0.5.16

### Minimal reproducible code

```rust
use winnow::{prelude::*, token::any};

/// Parse any two tokens
pub fn two_tokens(i: TokenSlice) -> PResult<()> {
    any.parse_next(i)?;
    any.parse_next(i)?;
    Ok(())
}

/// The stream is a vec of custom tokens.
type TokenSlice<'slice, 'input> = &'slice mut &'input [Token];

fn main() {
    // This parser should fail, because it's trying to parse two tokens and there's only one available.
    let one_token = [Token('a')];
    let result = two_tokens.parse(&one_token);

    let err = result.unwrap_err();
    let stream = err.input();

    // Expectation: I can find the bad token (the token which caused the error
    // by indexing err.offset() into err.input().
    // However, this actually gives an index-out-of-bounds panic, because
    // the offset is 1 and the input slice only has length 1.
    let bad_token = &stream[err.offset()];
    eprintln!("Bad token: {bad_token:?}");
}

#[derive(Debug, PartialEq, Eq, Clone)]
pub struct Token(char);
```

### Steps to reproduce the bug with the above code

cargo run

### Actual Behaviour

The `ParseError` reports error at offset 1, but the input stream is a slice of length 1, so 1 is not a valid index into the input.

### Expected Behaviour

Either report a valid index (in this case, 0 is the only valid index) or change the `offset()` method to return something different when the parser is trying to consume more tokens than the input stream contains. Maybe offset should return an enum:

```rust
enum ErrorCausedAt {
    /// The input stream had an unexpected or invalid token at this offset
    Offset(usize),
    /// The input ended while the parser was still trying to parse something
    UnexpectedEndOfInput,
}
```

As a Winnow user, my opinion is that returning ErrorCausedAt::UnexpectedEndOfInput is more helpful than returning the last index in the input. It demonstrates the problem more clearly, and forces me to consider these kind of edge cases when designing my application. In my case, I'm writing a parser for my programming language. This error would prompt me to output a better error message to my users, telling them they didn't finish writing their function properly.

### Additional Context

This is caused by the parser trying to consume two tokens from an input stream of length 1.

epage:
In particular, this calls out a problem when you conflate the two.

Fixes #341

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1182,6 +1182,9 @@ impl<I, E> ParseError<I, E> {
     }
 
     /// The location in [`ParseError::input`] where parsing failed
+    ///
+    /// **Note:** This is an offset, not an index, and may point to the end of input
+    /// (`input.len() + 1`) on eof errors.
     #[inline]
     pub fn offset(&self) -> usize {
         self.offset
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -1286,7 +1286,10 @@ where
 
 /// Useful functions to calculate the offset between slices and show a hexdump of a slice
 pub trait Offset<Start = Self> {
-    /// Offset between the first byte of `start` and the first byte of `self`
+    /// Offset between the first byte of `start` and the first byte of `self`a
+    ///
+    /// **Note:** This is an offset, not an index, and may point to the end of input
+    /// (`start.len() + 1`) when `self` is exhausted.
     fn offset_from(&self, start: &Start) -> usize;
 }
 


src/error.rs
```diff
@@ -1182,6 +1182,9 @@ impl<I, E> ParseError<I, E> {
     }
 
     /// The location in [`ParseError::input`] where parsing failed
+    ///
+    /// **Note:** This is an offset, not an index, and may point to the end of input
+    /// (`input.len() + 1`) on eof errors.
```
adamchalmers, line 1187: Actually, offset is `input.len()`on EOF errors.

PR #350, opened 2023-10-25
docs: fix list of parsers and combinators

vwkd:
Update the list of parsers and combinators to the new mutable input and `PResult` from #278.

Note, this was just dumb copy and paste. I didn't verify if it is correct.


epage:
Thanks!

PR #351, opened 2023-10-25
fix: typos

vwkd:
Fix some typos.


src/_tutorial/chapter_1.rs
```diff
@@ -19,7 +19,7 @@
 //! ```text
 //!                                    ┌─► Ok(what matched the parser)
 //!              ┌─────────┐           │
-//!  my input───►│my parser├──►either──┤
+//! my input───►│my parser├──►either──┤
```
epage, line 22: Is this still aligned correctly?

epage:
Thanks!

src/_tutorial/chapter_1.rs
vwkd: Ah, good catch! Saw it too late. Will have to post a follow up PR.

PR #358, opened 2023-10-30
fix: Deprecate signed Uint impls

epage:
Inspired by #355

### DIFF
--- a/src/ascii/mod.rs
+++ b/src/ascii/mod.rs
@@ -965,6 +965,7 @@ impl Uint for u128 {
     }
 }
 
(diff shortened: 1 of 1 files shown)


adriandelgado:
I think this should say `/// Deprecated since v0.5.17`.  On that subject, I personally find it useful to use parse an unsigned integer directly into a `i32`.The alternative is to use `dec_int` which has

adriandelgado:
@epage  Friendly ping to update the comment from `/// Deprecated since v0.15.17` to `/// Deprecated since v0.5.17`.

PR #362, opened 2023-11-01
feat(combinator): add separated

vwkd:
This attempts to add a `separated` parser that accepts a range similar to `repeat`.

Towards #98.

I tried to copy as much logic as possible from existing code. I'm unfamiliar with the code and may have missed some things, e.g. checkpoint resets, infinite loop checks, or correct error types. Please double check carefully and feel free to edit if it can be improved.

### DIFF
--- a/src/combinator/multi.rs
+++ b/src/combinator/multi.rs
@@ -513,7 +513,11 @@ where
     })
 }
 
-fn separated0_<I, O, C, O2, E, P, S>(parser: &mut P, sep: &mut S, i: &mut I) -> PResult<C, E>
+fn separated0_<I, O, C, O2, E, P, S>(
+    parser: &mut P,
+    separator: &mut S,
+    input: &mut I,
+) -> PResult<C, E>
 where
     I: Stream,
     C: Accumulate<O>,
@@ -521,46 +525,46 @@ where
     S: Parser<I, O2, E>,
     E: ParserError<I>,
 {
-    let mut res = C::initial(None);
+    let mut acc = C::initial(None);
 
-    let start = i.checkpoint();
-    match parser.parse_next(i) {
+    let start = input.checkpoint();
+    match parser.parse_next(input) {
         Err(ErrMode::Backtrack(_)) => {
-            i.reset(start);
-            return Ok(res);
+            input.reset(start);
+            return Ok(acc);
         }
         Err(e) => return Err(e),
         Ok(o) => {
-            res.accumulate(o);
+            acc.accumulate(o);
         }
     }
 
     loop {
-        let start = i.checkpoint();
-        let len = i.eof_offset();
-        match sep.parse_next(i) {
+        let start = input.checkpoint();
+        let len = input.eof_offset();
+        match separator.parse_next(input) {
             Err(ErrMode::Backtrack(_)) => {
-                i.reset(start);
-                return Ok(res);
+                input.reset(start);
+                return Ok(acc);
             }
             Err(e) => return Err(e),
             Ok(_) => {
                 // infinite loop check
-                if i.eof_offset() == len {
+                if input.eof_offset() == len {
                     return Err(ErrMode::assert(
-                        i,
+                        input,
                         "`separated` separator parser must always consume",
                     ));
                 }
 
-                match parser.parse_next(i) {
+                match parser.parse_next(input) {
                     Err(ErrMode::Backtrack(_)) => {
-                        i.reset(start);
-                        return Ok(res);
+                        input.reset(start);
+                        return Ok(acc);
                     }
                     Err(e) => return Err(e),
                     Ok(o) => {
-                        res.accumulate(o);
+                        acc.accumulate(o);
                     }
                 }
             }
@@ -615,7 +619,11 @@ where
     })
 }
 
-fn separated1_<I, O, C, O2, E, P, S>(parser: &mut P, sep: &mut S, i: &mut I) -> PResult<C, E>
+fn separated1_<I, O, C, O2, E, P, S>(
+    parser: &mut P,
+    separator: &mut S,
+    input: &mut I,
+) -> PResult<C, E>
 where
     I: Stream,
     C: Accumulate<O>,
@@ -623,42 +631,42 @@ where
     S: Parser<I, O2, E>,
     E: ParserError<I>,
 {
-    let mut res = C::initial(None);
+    let mut acc = C::initial(None);
 
     // Parse the first element
-    match parser.parse_next(i) {
+    match parser.parse_next(input) {
         Err(e) => return Err(e),
         Ok(o) => {
-            res.accumulate(o);
+            acc.accumulate(o);
         }
     }
 
     loop {
-        let start = i.checkpoint();
-        let len = i.eof_offset();
-        match sep.parse_next(i) {
+        let start = input.checkpoint();
+        let len = input.eof_offset();
+        match separator.parse_next(input) {
             Err(ErrMode::Backtrack(_)) => {
-                i.reset(start);
-                return Ok(res);
+                input.reset(start);
+                return Ok(acc);
             }
             Err(e) => return Err(e),
             Ok(_) => {
                 // infinite loop check
-                if i.eof_offset() == len {
+                if input.eof_offset() == len {
                     return Err(ErrMode::assert(
-                        i,
+                        input,
                         "`separated` separator parser must always consume",
                     ));
                 }
 
-                match parser.parse_next(i) {
+                match parser.parse_next(input) {
                     Err(ErrMode::Backtrack(_)) => {
-                        i.reset(start);
-                        return Ok(res);
+                        input.reset(start);
+                        return Ok(acc);
                     }
                     Err(e) => return Err(e),
                     Ok(o) => {
-                        res.accumulate(o);
+                        acc.accumulate(o);
                     }
                 }
             }
--- a/src/combinator/multi.rs
+++ b/src/combinator/multi.rs
@@ -342,6 +342,138 @@ where
 /// [`cut_err`][crate::combinator::cut_err].
 ///
 /// # Arguments
+/// * `m` The minimum number of iterations.
+/// * `n` The maximum number of iterations.
+/// * `parser` The parser that parses the elements of the list.
+/// * `sep` The parser that parses the separator between list elements.
+///
+/// # Example
+///
+/// Zero or more repetitions:
+/// ```rust
+/// # #[cfg(feature = "std")] {
+/// # use winnow::{error::ErrMode, error::ErrorKind, error::Needed};
+/// # use winnow::prelude::*;
+/// use winnow::combinator::separated;
+/// use winnow::token::tag;
+///
+/// fn parser(s: &str) -> IResult<&str, Vec<&str>> {
+///   separated(0.., "abc", "|").parse_peek(s)
+/// }
+///
+/// assert_eq!(parser("abc|abc|abc"), Ok(("", vec!["abc", "abc", "abc"])));
+/// assert_eq!(parser("abc123abc"), Ok(("123abc", vec!["abc"])));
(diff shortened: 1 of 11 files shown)


src/combinator/multi.rs
```diff
@@ -238,17 +238,8 @@ where
     let mut res = C::initial(Some(min));
     for count in 0..max {
         let start = input.checkpoint();
-        let len = input.eof_offset();
         match parse.parse_next(input) {
             Ok(value) => {
-                // infinite loop check: the parser must always consume
```
epage: Can you split this out into an issue to discuss?

My biggest concern for why I want to talk through this more is the inconsistent behavior for `repeat`, depending on what range is used

src/combinator/multi.rs
```diff
@@ -333,166 +333,367 @@ where
...
+/// assert_eq!(parser("abc123abc"), Ok(("123abc", vec!["abc"])));
+/// assert_eq!(parser("abc|def"), Ok(("|def", vec!["abc"])));
 /// assert_eq!(parser(""), Ok(("", vec![])));
 /// assert_eq!(parser("def|abc"), Ok(("def|abc", vec![])));
 /// # }
 /// ```
 #[doc(alias = "sep_by")]
+#[doc(alias = "sep_by1")]
 #[doc(alias = "separated_list0")]
-pub fn separated0<I, O, C, O2, E, P, S>(mut parser: P, mut sep: S) -> impl Parser<I, C, E>
```
epage, line 303: Can you keep `separated0` and just make it deprecated?

How I normally handle these transitions is I add the new feature in one commit, make the next commit update callers to it, and then the following commit deprecates it.  Keeps the commits small and buildable.

src/combinator/multi.rs
```diff
@@ -333,166 +333,367 @@ where
...
+    input: &mut I,
+) -> PResult<C, E>
+where
+    I: Stream,
+    C: Accumulate<O>,
+    P: Parser<I, O, E>,
+    S: Parser<I, O2, E>,
+    E: ParserError<I>,
+{
+    let mut acc = C::initial(None);
```
epage, line 468: To make reviews easier, I'd recommend renaming variables to a follow up commit

src/combinator/multi.rs
```diff
@@ -497,18 +497,13 @@ where
                     ));
                 }
 
-                let len = i.eof_offset();
```
epage: As this is removing extra work done compared to the original implementation, I'd recommend just squashing this commit into the prior commit

src/combinator/multi.rs
```diff
@@ -333,166 +333,367 @@ where
...
+
+    let start = input.checkpoint();
+    match parser.parse_next(input) {
+        Err(ErrMode::Backtrack(e)) => {
+            if min > 0 {
+                return Err(ErrMode::Backtrack(e.append(input, ErrorKind::Many)));
+            } else {
+                input.reset(start);
+                return Ok(acc);
+            }
```
epage, line 645: nit: swapping the cases so the condition is `min == 0` I feel would be cleaner / clearer

src/combinator/multi.rs
```diff
@@ -333,166 +333,367 @@ where
...
-                        }
-                        Err(e) => return Err(e),
-                        Ok(o) => {
-                            res.accumulate(o);
-                        }
-                    }
+    Ok(acc)
+}
+
+fn separated_m_n_<I, O, C, O2, E, P, S>(
```
epage, line 617: Can you add the infinite checks?

epage:
Thanks for moving this forward!

src/combinator/multi.rs
vwkd: Agree, split out into #366 with #365 to discuss it first.

src/combinator/multi.rs
vwkd: Done.

src/combinator/multi.rs
vwkd: True. Fixed now.

src/combinator/multi.rs
vwkd: The naming is consistent with the newer `repeat` functions. Since the existing `separated0` and `separated1` are now kept, the diff shouldn't be impacted anymore.

src/combinator/multi.rs
vwkd: Yes, I've added them back and moved the removal to #366.

src/combinator/multi.rs
```diff
@@ -504,6 +504,375 @@ where
...
+/// assert_eq!(parser("def|abc"), Ok(("def|abc", vec![])));
+/// # }
+/// ```
+#[doc(alias = "sep_by")]
+#[doc(alias = "sep_by1")]
+#[doc(alias = "separated_list0")]
+#[doc(alias = "separated_list1")]
+#[doc(alias = "separated_m_n")]
+#[inline(always)]
+pub fn separated<I, O, C, O2, E, P, S>(
```
epage, line 607: This commit should also includes tests for the new modes, matching the test cases covered for `repeat`

src/combinator/multi.rs
epage: Ideally the `separated0` and `separated1` would call into the new methods and the code would be arranged so the diffs would show that no behavior was changed in the bodies.  This also helps with doing `git blame`.  Like I said, later commits can do cleanups (much like your refactor PR for moving the functions, that was a byproduct of me doing the same thing and not getting to cleaning it up)

src/combinator/multi.rs
vwkd: Added. Reversal commit in #366.

src/combinator/multi.rs
```diff
@@ -333,166 +333,367 @@ where
...
-/// assert_eq!(parser("abc123abc"), Ok(("123abc", vec!["abc"])));
-/// assert_eq!(parser("abc|def"), Ok(("|def", vec!["abc"])));
-/// assert_eq!(parser(""), Err(ErrMode::Backtrack(InputError::new("", ErrorKind::Tag))));
-/// assert_eq!(parser("def|abc"), Err(ErrMode::Backtrack(InputError::new("def|abc", ErrorKind::Tag))));
-/// # }
-/// ```
-#[doc(alias = "sep_by1")]
-#[doc(alias = "separated_list1")]
-pub fn separated1<I, O, C, O2, E, P, S>(mut parser: P, mut sep: S) -> impl Parser<I, C, E>
+fn separated_n_<I, O, C, O2, E, P, S>(
```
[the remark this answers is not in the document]
vwkd: Added. Reversal commit in #366.

vwkd:
This should be good to go for another round.

The infinite loop checks for the element parser (not the separator parser) in `separated0_` currently make one test fail. Not sure what to do here.

epage:
> Hope this is now better and easier to review. Just to stress again, I'm not sure how correct this code is. Please review carefully.
> 
> Currently, the infinite loop checks for the element parser (not the separator parser) in `separated0_` make one test fail.

This PR should not be changing behavior and duplicating the `separated0` and `separated1` bodies, its hard to tell what behavior changed for us to talk about it.

vwkd:
I must have misunderstood the request to keep `separated0` and deprecating it.

As requested, I've now changed it back such that `seperated0` becomes `separated0_` (and similar for `1`), added tests, and moved the variable renames into a separate commit. The diff should be readable now. I've also removed again the infinite loop check for the element parser (not the separator parser), which was not in the existing implementation, which now makes all tests pass.

Please note, I kept the implementation and the updating call sites in separate commits as requested, but since the signature of the functions change, it can't be used at the call sites, which means the implementation commit (`feat(combinator): add separated`) won't compile.

Slightly off topic: Maybe it would make sense to adopt a squash/rebase flow instead of a merge flow, such that individual commits of the PR can be purely for ease of readability and review, since they don't need to compile as the whole PR gets squashed into a single commit on the main branch. A linear history and readable commit log is a nice bonus.

src/combinator/multi.rs
vwkd: Done.

epage:
> I must have misunderstood the request to keep `separated0` and deprecating it.
> 
> As requested, I've changed it back such that `seperated0` becomes `separated0_` (and similar for `1`). I've added tests, and moved the variable renames into a separate commit, such that the diff is more readable[1](#user-content-fn-1-6221bc56f380ac0f7f3903667bebb59e). I've again removed the extra infinite loop checks for the element parsers (not the existing one for the separator parsers) - which was not in the existing implementation - such that all tests pass.

This change should not be a breaking change.  You can both split out `separated0_` for reuse **and** keep `separator0`.  See #241.
 
> Slightly off topic: Maybe it would make sense to adopt a squash/rebase flow instead of a merge flow, such that individual commits of the PR can be purely for ease of readability and review, since they don't need to compile as the whole PR gets 

Squash/rebase workflows lose valuable history and intentionally sending someone broken commits to review can be confusing.

vwkd:
I think I understand the request now. I've kept deprecated `separated0` that reuse `separated0_` internally (similar for `1`) and a separated commit to remove them. This should keep each commit buildable.

src/combinator/multi.rs
```diff
@@ -468,51 +468,6 @@ where
...
-/// assert_eq!(parser("abc123abc"), Ok(("123abc", vec!["abc"])));
-/// assert_eq!(parser("abc|def"), Ok(("|def", vec!["abc"])));
-/// assert_eq!(parser(""), Ok(("", vec![])));
-/// assert_eq!(parser("def|abc"), Ok(("def|abc", vec![])));
-/// # }
-/// ```
-#[doc(alias = "sep_by")]
-#[doc(alias = "separated_list0")]
-#[deprecated]
-pub fn separated0<I, O, C, O2, E, P, S>(mut parser: P, mut sep: S) -> impl Parser<I, C, E>
```
epage, line 370: One of my pieces of guidance for this PR is that there should be no breaking changes.  Please drop this commit from the PR.

src/combinator/multi.rs
```diff
@@ -463,45 +611,218 @@ where
...
-                        Err(e) => return Err(e),
-                        Ok(o) => {
-                            res.accumulate(o);
+                    }
+                    Err(e) => {
+                        return Err(e);
+                    }
+                    Ok(o) => {
+                        // infinite loop check
+                        if input.eof_offset() == len {
```
epage, line 727: Infinite loop checks only go on the separator

src/combinator/multi.rs
```diff
@@ -454,6 +601,7 @@ where
 /// ```
 #[doc(alias = "sep_by1")]
 #[doc(alias = "separated_list1")]
+#[deprecated]
```
epage: Don't forget to list the `since` and `note`

epage:
Everything should be good to go after these last couple of comments are resolved!  Thanks for your work on this!

src/combinator/multi.rs
vwkd: Could you specify what values you desire?

vwkd: Done. Assumed the next version will be `0.5.19`.

src/combinator/multi.rs
vwkd: Done.

src/combinator/multi.rs
vwkd: Removed.

vwkd:
Addressed the remaining comments.

To lessen the previously breaking change, before I had considered proposing this in a follow up. But since this is now a non-breaking additive change, we could take the opportunity to choose the right name now which would also avoid another breaking change in the future. I think naming it `separate` (verb, instead of adjective `separated`) fits better and also aligns with the existing `repeat`.

epage:
If you'd like to discuss name changes, I'd recommend looking into whats been done before on https://github.com/winnow-rs/winnow/discussions/95 and posting your idea there.

epage:
Thanks!

Issue #365, opened 2023-11-01
When do we need infinite loop checks?


vwkd:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### rust version

rustc 1.72.1

### winnow version

v0.5.18

### Minimal reproducible code

While trying to implement a `separated` parser that accepts a range in #362, I found some [combinator parsers](https://github.com/winnow-rs/winnow/blob/main/src/combinator/multi.rs) have "infinite loop checks" like this

```rs
if input.eof_offset() == len {
    return Err(ErrMode::assert(i, "sep parsers must always consume"));
}
```

It would be good to know if and when they are necessary. For example

- are they only necessary in infinite loops or also in finite loops?
- are they necessary for each parser when there are multiple parsers, e.g. element parser and separator parser in `separated`

epage:
They are needed to prevent infinite loops.  The odd situation is when a ranged combinator can sometimes be at risk to an infinite loop and sometimes not.  I think consistency is better as it makes the behavior clearer (no weird caveats in documentation).

`repeat(5, ...)` does not implement the check.  I believe `repeat(1..5, ...)` does implement the check.

vwkd:
Expanding on my initial questions:

- Are the checks needed for `separated` at all assuming the element parser and separator parser are different (otherwise it would just be like a `repeat`)[^1]?
- Are the checks necessary for each parser when there are multiple parsers, e.g. element parser and separator parser in separated
- Are the checks also necessary in finite loops?

[^1]: Maybe we should verify the element parser and separator parser passed to `separated` are actually different…)

epage:
Yes, the checks are needed for `separated`.  If you have two parsers that can each sometimes match an empty string, there could be cases where they both match an empty string and you get in an infinite loop

We only need the check for sep or regular parser.  We don't need it for both.  Its very common to have an empty element between separators, so we bias towards that in the implementation.  See also #325.

I spoke to the checks for finite ranges earlier.

vwkd:
Makes sense.

[...]

PR #366, opened 2023-11-01
fix(combinator): remove infinite loop check in finite loop

vwkd:
From #365, this proposes to remove the infinite loop check in the finite loop in `separated_n_` added in #362.


vwkd:
Closing in favor of adding the infinite loop check to `repeat_n_` in #368 as discussed in https://github.com/winnow-rs/winnow/issues/365#issuecomment-1793243185

PR #368, opened 2023-11-04
fix(combinator)!: add missing infinite loop check to repeat

vwkd:
Add the missing infinite loop check to `repeat_n_` to make the error consistent.

As discussed in #365.

### DIFF
--- a/src/combinator/multi.rs
+++ b/src/combinator/multi.rs
@@ -211,8 +211,14 @@ where
     let mut res = C::initial(Some(count));
 
     for _ in 0..count {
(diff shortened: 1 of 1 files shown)


epage:
Thanks for doing this follow up!

As this is a breaking change, I lean towards waiting for enough of these or something more pressing to push us to breaking compatibility.  I'm trying to hold off and consolidate on those to see how well we can evolve and manage things without breaking to get an idea of when 1.0 might be appropriate.

epage:
I merged this as part of #449

PR #370, opened 2023-11-04
docs: Separated takes a range, not separate m and n arguments

VorpalBlade:
The combination `separated` takes a range, not separate m and n arguments. Fix the documentation.


epage:
Thanks!

PR #405, opened 2023-12-27
perf(token): Allow optimizer remove branch

### LINKED ISSUE #404: Any combinator function the if-else logic in this code appears to have no practical difference

### winnow version

0.5.30

### Describe your use case

The **if-else** logic in this code appears to have no practical difference. In both the if and else branches, the **any_::<_, _, true>(input)** function is called. Consequently, the if-else logic might be redundant.

> https://github.com/winnow-rs/winnow/blob/02561943e396aab9fc268ec00c88273e3315ef0a/src/token/mod.rs#L46-L59

```diff
pub fn any<I, E: ParserError<I>>(input: &mut I) -> PResult<<I as Stream>::Token, E>
where
    I: StreamIsPartial,
    I: Stream,
{
    trace("any", move |input: &mut I| {
        if <I as StreamIsPartial>::is_partial_supported() {
+            any_::<_, _, true>(input)
+        } else {
+            any_::<_, _, true>(input)
        }
    })
    .parse_next(input)
}
```

The if-else logic might be redundant and can be simplified by directly calling `any_::<_, _, true>(input)`.

_No response_

### Additional Context

_No response_

epage:
Not entirely clear if its needed in this case but I generally have found that these const conditionals seem to provide more optimized results.

I did a `rg is_partial_support -A 4` to see if any others had this problem and this seemed to be isolated.

Fixes #404

### DIFF
--- a/src/token/mod.rs
+++ b/src/token/mod.rs
@@ -52,7 +52,7 @@ where
         if <I as StreamIsPartial>::is_partial_supported() {
             any_::<_, _, true>(input)
         } else {
-            any_::<_, _, true>(input)
+            any_::<_, _, false>(input)
         }
     })
     .parse_next(input)


baoyachi:
@epage Out of curiosity, what is 'rg is_partial_support -A 4'?

DJDuque:
> Out of curiosity, what is 'rg is_partial_support -A 4'?

https://github.com/BurntSushi/ripgrep

epage:
We have an always-inlined function that converts the function call `is_partial_supported` to a monomorphized version of the parser.  This makes the optimizer a lot more likely to detect code as dead and remove it, reducing binary size and improving performance.

I searched for `-A4` to show me the 4 lines after the call to `is_partial_supported` as generally it follows a straight `if/else` pattern that will show up directly in the console.  If it isn't so direct, the pattern can raise my attention to directly check in the source.

Issue #409, opened 2024-01-01
Location aware error context


urso:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### winnow version

0.5.31

### Describe your use case

Using ContextError in combination with backtracking via `alt` can lead to confusing error reports. This is because `ContextError` basically discards the current context on `or`.

In presence of backtracking ideally I would wish for a context from the longest matching parser only. If 2 parsers did error at the same position the contexts should be merged. 

### Describe the solution you'd like

Provide a `LocationContextError` that I can use in case of `ContextError`. The `LocationContextError` would track the longest parsed location. This requires that the input implements `Location`.
When merging contexts via `AddContext` or when adding some context to the error, the `LocationContextError` should discard contexts at lower positions.

### Alternatives, if applicable

Encourage users to implement their own location aware context error. I have 2 implementations by now, but it would be nice to share the code in order to make it more generally available.

epage:
I feel there is something missing in this idea because of the focus on other location types.

`Location` is intended to give an absolute position for the current location and isn't sufficient on its own to make a span as you need two locations.

I'm also concerned this couples the ideas of Location tracking in errors and backtracking behavior.

As one alternative, we could have an Error decorator that tracks a `Ord` `Checkpoint` and handles the backtracking case.

urso:
> I feel there is something missing in this idea because of the focus on other location types.
>
> Location is intended to give an absolute position for the current location and isn't sufficient on its own to make a span as you need two locations.

Sounds fair. Indeed it was not my intention to add location tracking (although that would might a nice addition as well). Still in presence of backtracking I find the current `ContextError` lacking.

For example in a parser like this (SQL dialect):

```
alt((
  parse_create_table,
  parse_create_view,
  parse_drop_table
))
```

With context error (without splicing in `cut_err` or using `dispatch`) the `ContextError` will focus on `parse_drop_table`, which leads to confusing error messages.

> As one alternative, we could have an Error decorator that tracks a Ord Checkpoint and handles the backtracking case.

I wonder how this will look like.

Related to #410: 
[...]

PR #410, opened 2024-01-01
Location aware error context

urso:
Enhancement Request: https://github.com/winnow-rs/winnow/issues/409

This PR implements the `LocationContextError`. The implementation follows the `ContextError` type:

```
pub struct LocationContextError<L = usize, C = StrContext> {
   ...
}
```

Main differences to `ContextError`:
- we add a `pos` field to the `LocationContextError`.
- `ParserError.or` will choose the context based on the longest parse. We merge the contexts into a common `Vec` in case both errors did reach the same position in the input stream
- `add_context` will discard context details based on the input location

Although `Location.location` returns an `usize`, this type is not always appropriate in all situations. For example in one of my parsers I have a lexer phase and some parse steps that just recognize a sequence of tokens. The actual parsing happens later. For that reason I'm having a `struct Span(Range<u32>)` as location.

In order to make the location type configurable we allow users to set the location type via `L`.

Note: I didn't want to modify the `Location` trait to make the output type configurable yet. Although I think that would improve the usability of `LocationContextError`. This is why `ParserError` and `AddContext` are implemented for `usize` only. But their implementations can easily be copied for custom location types. The helper methods to merge/add context information only require `L: Ord`.

Testing: 
I didn't find tests for the other error types. So I didn't add new ones. I leave this PR in draft mode for now and will apply the ContextError to one of my projects that uses winnow as lexer and as token based parser. In that project I have location types `u32` and `struct Span(Range<u32>)`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -17,12 +17,14 @@
 //! - [`ErrorKind`]
 //! - [`InputError`] (mostly for testing)
 //! - [`ContextError`]
+//! - [`LocationContextError`] collect context from longest match
 //! - [`TreeError`] (mostly for testing)
 //! - [Custom errors][crate::_topic::error]
 
 #[cfg(feature = "alloc")]
 use crate::lib::std::borrow::ToOwned;
 use crate::lib::std::fmt;
+use crate::stream::Location;
 use core::num::NonZeroUsize;
 
 use crate::stream::AsBStr;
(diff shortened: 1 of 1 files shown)


urso:
Update:
I removed the `L` location type parameter from `LocationContextError`. We store the position as `usize` and require the input stream to implement `Location`.

Issue #411, opened 2024-01-02
`winnow::ascii::float` doesn't match `+inf` or `-inf`


dezyh:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### rust version

rustc 1.75.0 (82e1608df 2023-12-21)

### winnow version

0.5.31

### Minimal reproducible code
Link to [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5d040e513c96c96b10d1976093754516
)
```rust

#[cfg(test)]
mod test {
    use winnow::error::ContextError;
        
    #[test]
    fn test_std_parse_inf_camel() {
        assert_eq!(
            "Inf".parse::<f64>().unwrap(), 
            f64::INFINITY,
        );
    }

    #[test]
    fn test_std_parse_pos_inf_camel() {
        assert_eq!(
            "+Inf".parse::<f64>().unwrap(), 
            f64::INFINITY,
        );
    }
    
    #[test]
    fn test_std_parse_neg_inf_camel() {
        assert_eq!(
            "-Inf".parse::<f64>().unwrap(), 
            f64::NEG_INFINITY,
        );
    }
    
    #[test]
    fn test_std_parse_inf_lower() {
        assert_eq!(
            "inf".parse::<f64>().unwrap(), 
            f64::INFINITY,
        );
    }

    #[test]
    fn test_std_parse_pos_inf_lower() {
        assert_eq!(
            "+inf".parse::<f64>().unwrap(), 
            f64::INFINITY,
        );
    }
    
    #[test]
    fn test_std_parse_neg_inf_lower() {
        assert_eq!(
            "-inf".parse::<f64>().unwrap(), 
            f64::NEG_INFINITY,
        );
    }
        
    #[test]
    fn test_winnow_parse_inf_camel() {
        assert_eq!(
            winnow::ascii::float::<&str, f64, ContextError>(&mut "Inf").unwrap(), 
            f64::INFINITY,
        );
    }

    #[test]
    fn test_winnow_parse_pos_inf_camel() {
        assert_eq!(
            winnow::ascii::float::<&str, f64, ContextError>(&mut "+Inf").unwrap(), 
[...]

PR #412, opened 2024-01-02
Fixes `winnow::ascii::float` not matching `+inf` or `-inf`

dezyh:
Fixes `winnow::ascii::float` not matching any variants of `inf`/`infinite` prefixed by a `+`/`-` sign as mentioned in https://github.com/winnow-rs/winnow/issues/411.

### DIFF
--- a/src/ascii/mod.rs
+++ b/src/ascii/mod.rs
@@ -1367,8 +1367,12 @@ where
     alt((
         recognize_float,
         crate::token::tag_no_case("nan"),
-        crate::token::tag_no_case("infinity"),
-        crate::token::tag_no_case("inf"),
+        (
+            opt(one_of(['+', '-'])),
+            crate::token::tag_no_case("infinity"),
+        )
+            .recognize(),
+        (opt(one_of(['+', '-'])), crate::token::tag_no_case("inf")).recognize(),
     ))
     .parse_next(input)
 }
--- a/src/ascii/tests.rs
+++ b/src/ascii/tests.rs
@@ -566,24 +566,17 @@ mod complete {
 
             println!("now parsing: {} -> {}", test, expected32);
 
-            if (test.starts_with('-') || test.starts_with('+')) && expected32.is_infinite() {
-                assert!(float::<_, f32, ()>.parse_peek(test.as_bytes()).is_err());
-                assert!(float::<_, f32, ()>.parse_peek(test).is_err());
-                assert!(float::<_, f64, ()>.parse_peek(test.as_bytes()).is_err());
-                assert!(float::<_, f64, ()>.parse_peek(test).is_err());
-            } else {
-                assert_parse!(
-                    float.parse_peek(test.as_bytes()),
-                    Ok((&b""[..], expected32))
-                );
-                assert_parse!(float.parse_peek(test), Ok(("", expected32)));
+            assert_parse!(
+                float.parse_peek(test.as_bytes()),
+                Ok((&b""[..], expected32))
+            );
+            assert_parse!(float.parse_peek(test), Ok(("", expected32)));
 
-                assert_parse!(
-                    float.parse_peek(test.as_bytes()),
-                    Ok((&b""[..], expected64))
-                );
-                assert_parse!(float.parse_peek(test), Ok(("", expected64)));
-            }
+            assert_parse!(
+                float.parse_peek(test.as_bytes()),
+                Ok((&b""[..], expected64))
+            );
+            assert_parse!(float.parse_peek(test), Ok(("", expected64)));
         }
 
         let remaining_exponent = "-1.234E-";
--- a/src/ascii/tests.rs
+++ b/src/ascii/tests.rs
@@ -592,9 +592,27 @@ mod complete {
             Err(ErrMode::Cut(InputError::new("", ErrorKind::Slice)))
         );
 
-        let (i, nan) = float::<_, f32, ()>.parse_peek("NaN").unwrap();
-        assert!(nan.is_nan());
-        assert_eq!(i, "");
+        let nan_test_cases = ["nan", "NaN", "NAN"];
+
+        for test in nan_test_cases {
+            println!("now parsing: {}", test);
+
+            let (remaining, parsed) = float::<_, f32, ()>.parse_peek(test.as_bytes()).unwrap();
+            assert!(parsed.is_nan());
+            assert!(remaining.is_empty());
+
+            let (remaining, parsed) = float::<_, f32, ()>.parse_peek(test).unwrap();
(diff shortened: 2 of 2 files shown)


dezyh:
Aside, I saw this `cargo clippy` lint near my changes
```
warning: use of deprecated function `token::tag_no_case`: Replaced with `tag(ascii::Caseless(_))`
  --> src/ascii/mod.rs:18:19
   |
18 | use crate::token::tag_no_case;
   |                   ^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

However, proposed solution does not work in this method so I'm not considering that lint with this PR. However, if you have any advice, let me know and I'll take a quick look.

This was the error when using `crate::token::tag` and `crate::ascii::Caseless`
```
error[E0277]: the trait bound `I: Compare<Caseless<&str>>` is not satisfied
    --> src/ascii/mod.rs:1370:9
     |
1370 |         tag(Caseless("nan")),
     |         ^^^ the trait `Compare<Caseless<&str>>` is not implemented for `I`
     |
note: required by a bound in `token::tag`
    --> src/token/mod.rs:141:17
     |
138  | pub fn tag<T, I, Error: ParserError<I>>(tag: T) -> impl Parser<I, <I as Stream>::Slice, Error>
     |        --- required by a bound in this function
...
141  |     I: Stream + Compare<T>,
     |                 ^^^^^^^^^^ required by this bound in `tag`
help: consider further restricting this bound
     |
1361 |     I: StreamIsPartial + stream::Compare<ascii::Caseless<&str>>,
     |                        ++++++++++++++++++++++++++++++++++++++++

error[E0277]: the trait bound `I: Compare<Caseless<&str>>` is not satisfied
    --> src/ascii/mod.rs:1371:35
     |
1371 |         (opt(one_of(['+', '-'])), tag(Caseless("infinity"))).recognize(),
     |                                   ^^^ the trait `Compare<Caseless<&str>>` is not implemented for `I`
     |
note: required by a bound in `token::tag`
    --> src/token/mod.rs:141:17
     |
138  | pub fn tag<T, I, Error: ParserError<I>>(tag: T) -> impl Parser<I, <I as Stream>::Slice, Error>
     |        --- required by a bound in this function
...
141  |     I: Stream + Compare<T>,
     |                 ^^^^^^^^^^ required by this bound in `tag`
help: consider further restricting this bound
     |
1361 |     I: StreamIsPartial + stream::Compare<ascii::Caseless<&str>>,
     |                        ++++++++++++++++++++++++++++++++++++++++

error[E0277]: the trait bound `I: Compare<Caseless<&str>>` is not satisfied
    --> src/ascii/mod.rs:1372:35
     |
1372 |         (opt(one_of(['+', '-'])), tag(Caseless("inf"))).recognize(),
     |                                   ^^^ the trait `Compare<Caseless<&str>>` is not implemented for `I`
     |
note: required by a bound in `token::tag`
    --> src/token/mod.rs:141:17
     |
138  | pub fn tag<T, I, Error: ParserError<I>>(tag: T) -> impl Parser<I, <I as Stream>::Slice, Error>
     |        --- required by a bound in this function
...
141  |     I: Stream + Compare<T>,
     |                 ^^^^^^^^^^ required by this bound in `tag`
help: consider further restricting this bound
     |
1361 |     I: StreamIsPartial + stream::Compare<ascii::Caseless<&str>>,
     |                        ++++++++++++++++++++++++++++++++++++++++
```

dezyh:
In retrospect, I noticed the infinite float values are actually comparable to themselves (unlike `NaN`), so we could move all the infinity variants into the main `test_cases` which makes my refactoring changes a little redundant. 

Let me know what you think.

src/ascii/tests.rs
```diff
@@ -575,16 +575,31 @@ mod complete {
...
-        assert!(nan.is_nan());
-        assert_eq!(i, "");
-
-        let (i, inf) = float::<_, f32, ()>.parse_peek("inf").unwrap();
-        assert!(inf.is_infinite());
-        assert_eq!(i, "");
-        let (i, inf) = float::<_, f32, ()>.parse_peek("infinity").unwrap();
-        assert!(inf.is_infinite());
-        assert_eq!(i, "");
+        let mut special_test_cases = vec!["NaN", "inf", "infinity"];
```
epage: nit: this could be a `[]` instead of `vec![]`

src/ascii/tests.rs
```diff
@@ -575,7 +575,22 @@ mod complete {
...
+            "Infinity",
+            "-inf",
+            "-Inf",
+            "-infinity",
+            "-Infinity",
+            "+inf",
+            "+Inf",
+            "+infinity",
+            "+Infinity",
+        ];
```
epage, line 593: A commit should not be failing tests.  What I would expect is the commit should pass tests and the follow up commit should update the tests to show they still work.

Benefits:
- Helps test the tests, making sure you are testing the right thing
- Makes the intended behavior change clearer

src/ascii/mod.rs
```diff
@@ -15,6 +15,7 @@ use crate::error::{ErrMode, ErrorKind, Needed};
 use crate::stream::{AsBStr, AsChar, ParseSlice, Stream, StreamIsPartial};
 use crate::stream::{Compare, CompareResult};
 use crate::token::one_of;
+use crate::token::tag_no_case;
```
epage, line 18: nit: this is an unrelated change, so please revert it

(notice that the function has an `allow(deprecated)` on it for `tag_no_case` because we can't resolve the deprecation until we make a breaking change)

epage:
> In retrospect, I noticed the [infinite float values](https://github.com/winnow-rs/winnow/pull/412/files#diff-be1c8124f4111e9a2aa0e42369d9cc208120284851786b20f09e63aa47077383R581-R592) are actually comparable to themselves (_"Infinite operands of the same sign shall compare equal"_) (unlike `NaN`), so we could move all the infinity variants into the main `test_cases` which makes my refactoring changes a little redundant.
> 
> Let me know what you think.

I can go either way on this

src/ascii/tests.rs
dezyh: Thanks. I saw this too but didn't want to add unnecessary changes to the PR. 
I've changed both this usage and the existing usage of `vec![]` to an array `[]`.

src/ascii/mod.rs
dezyh: Apologies, I didn't realize it was the `use crate::token::tag_no_case` which was triggering the clippy deprecated lint (rather than simply calling `crate::token::tag_no_case`).

src/ascii/tests.rs
epage: Ah, I had missed that `vec!` was already used

PR #415, opened 2024-01-03
Longest

urso:
Continuation of #413 and #410

This PR introduces the `MergeContext` trait, so we can call `merge_context` during `or`.


src/error.rs
```diff
@@ -1,4 +1,4 @@
-//! # Error management
+//! # Errormanagement
```
epage: Please clean up the commit history
- This commit shouldn't have this change
- Most other content from this commit should likely be squashed into a previous commit
- Replace the `AddContext::...` commits with the `MergeContext` trait commit which will hopefully remove the roundtripping of `ParseError` getting a `C` parameter and then dropping it
- Adding of tests is unrelated to the merge context commit and should either be in the `LongestMatch` commit or its own commit

src/error/tests.rs
```diff
@@ -0,0 +1,70 @@
...
+    use crate::{
+        combinator::{alt, eof, terminated},
+        Located,
+    };
+
+    type Input<'a> = Located<&'a str>;
+    type Error<'a> = LongestMatch<Input<'a>, ContextError<&'static str>>;
+    type PResult<'a, O> = crate::error::PResult<O, Error<'a>>;
+
+    fn tag<'a>(t: &'static str) -> impl Parser<Input<'a>, &'a str, Error<'a>> {
```
epage, line 14: Why are we adding a hand-written `tag` rather than just using the fact that `&'static str` implements `Parser`?

src/error/tests.rs
```diff
@@ -0,0 +1,70 @@
+use super::*;
+
+mod longest_match {
+    use super::*;
+    use crate::{
+        combinator::{alt, eof, terminated},
+        Located,
+    };
+
+    type Input<'a> = Located<&'a str>;
```
epage, line 10: I feel like we should simplify these tests by removing `Located`

src/error/tests.rs
```diff
@@ -0,0 +1,70 @@
...
+        err.context().copied().collect::<Vec<_>>()
+    }
+
+    #[test]
+    fn single_longest_match_first_in_alt() {
+        fn parser<'a>(input: &mut Input<'a>) -> PResult<'a, &'a str> {
+            alt((syms("abcd"), syms("abc"), syms("ab"))).parse_next(input)
+        }
+        let ctx = expect_err(parser(&mut Input::new("abcde")));
+        assert_eq!(ctx, vec!["abcd"]);
```
epage, line 41: imo we should also have a test that verifies that we get the expected rendering of these errors.

src/error/tests.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        let ctx = expect_err(parser(&mut Input::new("abd")));
+        assert_eq!(ctx, vec!["abcd", "abc"]);
+
+        let ctx = expect_err(parser(&mut Input::new("deg")));
+        assert_eq!(ctx, vec!["def", "defg"]);
+    }
+
+    #[test]
+    #[ignore]
```
epage, line 58: Why is this test ignored?

src/error.rs
urso: Sure, will clean up the commits before undrafting the PR.

src/error/tests.rs
urso: When using tag or `&;static str` then winnow uses a specialized compare to check if the inputs prefix matches that string.. That is on backtracking the checkpoints all point to the same location. Alternatively I maybe could have used `('a', 'b', 'c')` as test input.

epage: I see.  Yeah, I'd prefer something else like a tuple or array so we don't have to deal with the correctness of custom parser implementations.

src/error/tests.rs
urso: The test is failing because. That one shows the actual problem we face when using `Ord` on `Checkpoint`. 

For context:

> Unfortunately I think `Ord` is not the right trait. I also added some tests to see if it works as expected.
The problem becomes obvious when having partial input. In that case we have a parser that has consumed all input, but did produce an error and a second parser that did fail early. In that case we are comparing `""` with `"rest"`. Apparently `"rest" > ""`, which makes us report the wrong context.
>
> In case of `&'a str` as input [`SliceOrd`](https://doc.rust-lang.org/std/primitive.slice.html#impl-Ord-for-%5BT%5D) will be used. Stepping through with the debugger it looks like [this implementation is used](https://doc.rust-lang.org/src/core/slice/cmp.rs.html#179).

src/error.rs
```diff
@@ -313,6 +334,19 @@ pub trait AddContext<I, C = &'static str>: Sized {
...
+pub trait MergeContext: Sized {
+    /// Apply the context from `other` into `self`
+    #[inline]
+    fn merge_context(self, _other: Self) -> Self {
+        self
+    }
+
+    /// Remove all context
+    #[inline]
+    fn clear_context(&mut self) {}
```
epage, line 347: As this is a new trait and if you implement one of these, you likely need to implement the other, should we drop the inherent implementations?

src/error/tests.rs
urso: Introducing AsOrd did fix the test.

urso:
Made some changes like introducing the MergeContext trait and added a few more tests.

Also update the commit history.

src/stream/mod.rs
```diff
@@ -1466,6 +1466,31 @@ where
     }
 }
 
+/// Used to compare checkpoints
+pub trait AsOrd {
+    /// The type used to compare checkpoint positions
+    type Ord: Ord + Clone + core::cmp::Ord + crate::lib::std::fmt::Debug;
```
epage, line 1472: Isn't `Ord + core::cmp::Ord` redundant?

src/stream/mod.rs
```diff
@@ -2225,6 +2225,22 @@ where
 #[derive(Copy, Clone, Debug)]
 pub struct Checkpoint<T>(T);
 
+impl<T: PartialEq> PartialEq for Checkpoint<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.eq(&other.0)
+    }
+}
+
+impl<T: Eq> Eq for Checkpoint<T> {}
```
epage, line 2234: Do we still need these?

src/error.rs
```diff
@@ -536,6 +560,32 @@ impl<C, I> AddContext<I, C> for ContextError<C> {
...
+        // self and other get consumed to produce the new Context error.
+        // We choose the vector with the larger capacity to reduce the chance of reallocations.
+        #[cfg(feature = "alloc")]
+        {
+            let (mut context, other) = if self.context.capacity() >= other.context.capacity() {
+                (self.context, other.context)
+            } else {
+                (other.context, self.context)
+            };
+            context.extend(other);
```
epage, line 579: I feel like this could lead to confusing ordering.

Should we instead check if one is empty and instead pick the other?

src/stream/mod.rs
epage, line 1472: For `Clone` and `Debug`, if this is because `LongestMatch` impls those traits, wouldn't that be dependent on whether the `Ord` does so, meaning we don't need these bounds?

src/error/mod.rs
```diff
@@ -745,6 +749,194 @@ impl crate::lib::std::fmt::Display for StrContextValue {
...
+    E: MergeContext + Default,
+{
+    /// Create an empty error
+    pub fn new() -> Self {
+        Self {
+            checkpoint: None,
+            inner: Default::default(),
+        }
+    }
+}
```
epage, line 777: Why do we offer an empty error?

src/error/mod.rs
```diff
@@ -745,6 +749,194 @@ impl crate::lib::std::fmt::Display for StrContextValue {
...
+impl<I, E> core::cmp::PartialEq for LongestMatch<I, E>
+where
+    I: Stream,
+    <I as Stream>::Checkpoint: AsOrd + core::cmp::PartialEq,
+    E: MergeContext + core::cmp::PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.checkpoint == other.checkpoint && self.inner == other.inner
+    }
+}
```
epage, line 789: Do we need this?  `ContextError` needs it because its the default and so we need this for testing parsers.  Here, we likely shouldn't be doing equality check tests

src/error/mod.rs
epage: And is this the only reason `checkpoint` is an `Option`?

src/error/mod.rs
```diff
@@ -745,6 +749,194 @@ impl crate::lib::std::fmt::Display for StrContextValue {
...
+                self.inner = self.inner.append(input, kind);
+            }
+            core::cmp::Ordering::Greater => {}
+        }
+        self
+    }
+
+    #[inline]
+    fn or(self, other: Self) -> Self {
+        self.merge_context(other)
```
epage, line 862: Shouldn't this be checking the checkpoints?

src/error/mod.rs
```diff
@@ -745,6 +749,194 @@ impl crate::lib::std::fmt::Display for StrContextValue {
...
+        Self {
+            checkpoint: Some(input.checkpoint()),
+            inner: E::from_error_kind(input, kind),
+        }
+    }
+
+    #[inline]
+    fn append(mut self, input: &I, kind: ErrorKind) -> Self {
+        let checkpoint = input.checkpoint();
+        match self.cmp_with_checkpoint(&checkpoint) {
```
epage, line 847: When is this needed?

src/error/mod.rs
epage: Oh, right, thats in `merge_context

src/error/mod.rs
```diff
@@ -745,6 +749,194 @@ impl crate::lib::std::fmt::Display for StrContextValue {
...
+            core::cmp::Ordering::Equal => {
+                self.inner = self.inner.merge_context(other.inner);
+                self
+            }
+        }
+    }
+
+    fn clear_context(self) -> Self {
+        Self {
+            checkpoint: None,
```
epage, line 908: Why do we need to reset it when clearing?

src/error/tests.rs
```diff
@@ -0,0 +1,190 @@
...
+        label: &'static str,
+    ) -> impl Parser<&'a str, &'a str, Error<'a>> {
+        (p, eof).recognize().context(label)
+    }
+
+    #[test]
+    fn parser_error_implementation() {
+        let input = "abcd";
+        let checkpoint1 = &&input[2..];
+        let checkpoint2 = &&input[3..];
```
epage, line 20: If we keep these tests, we should name these checkpoints to make reading the asserts clearer

src/error/tests.rs
```diff
@@ -0,0 +1,190 @@
+use super::*;
+
+mod longest_match {
```
epage, line 3: My primary concern with these tests is they focus on all of the details but they don't make sure the details are actually correct for which we'd need tests that render error messages.

src/stream/mod.rs
urso: Oops, some oversight when combining the commits into one.

src/stream/mod.rs
urso: These are currently used by the tests. We can update the tests to compare the `err.into_inner()` or compare via `format!(...)`.

src/error.rs
urso: I didn't really care about order if there are multiple parsers with the same longest match, but I see how this might be confusing to developers or mess eventually mess with tests when they change parsers..

> Should we instead check if one is empty and instead pick the other?

👍

src/error/mod.rs
urso: The reason for `checkpoint` to be `Option` is the `LongestMatch` currently implements the `MergeContext` trait as well. When calling `clear_context` we must set `checkpoint: None`, such that the empty context is always `<` then any non-empty context. Otherwise merging 2 context would give us some inconsistent output.

I'm removing `impl MergeContext for LongestMatch`, then we should get rid of the `Option` for `checkpoint`.

src/error/mod.rs
urso: These are used by the tests.

As `LongestMatch` is no error on itself itself, but a decorator most likely used with `ContextError` I did feel like being able to compare `LongestMatch<ContextError>` would be the right thing to have.

src/error/mod.rs
urso: See my comment here https://github.com/winnow-rs/winnow/pull/415#discussion_r1443832350

I'm removing `impl MergeContext` or `LongestMatch`. This also removes the need or `checkpoint` to be `Option`.

src/error/tests.rs
urso: By "rendered error message" you mean using `format!(...)` ?

I didn't do this because `LongestMatch` does not directly implement any rendering itself. It will just delegates rendering to the inner error. That is the test output would depend on the current implementation of `ContextError`.

As LongestMatch acts as a 'guard' on the wrapped error when merging/adding context details I wanted to make sure with the tests that this 'guard' works as expected.

src/stream/mod.rs
urso: We can remove them. `Clone` was required because I initially stored the `ord` in `LongestMatch`. I updated LongestMatch to store the actual checkpoint now.

src/error/tests.rs
urso: I just noticed that `ErrMode` uses `Debug` internally to render the error.

Is this what you have had in mind?

```
    #[test]
    fn multi_longest_match() {
        let mut parser = alt((
            pattern(('d', 'e', 'f'), "don't want"),
            pattern(('a', 'b', 'c', 'd'), "wanted 1"),
            pattern(('a', 'b', 'c'), "wanted 2"),
            pattern(('d', 'e', 'f', 'g'), "don't want"),
        ));

        let mut input = "abd";
        let checkpoint = &&input[2..]; // 2 characters consumed by longest match
                                       //
        assert_eq!(
            format!("{}", parser.parse_next(&mut input).unwrap_err(),),
            r#"Parsing Error: LongestMatch { checkpoint: Checkpoint("d"), inner: ContextError { context: ["wanted 1", "wanted 2"], cause: None } }"#,
        );
    }

```

epage: > I didn't do this because LongestMatch does not directly implement any rendering itself. It will just delegates rendering to the inner error. That is the test output would depend on the current implementation of ContextError.

For me, without a `Display` test (not `Debug` or the current state of context), we aren't showing that we've been able to achieve what this is trying to accomplish.

axelkar:
Any updates on this?

PR #429, opened 2024-01-20
Fix link to `dec_int` in the docs

nvbn:
At the moment in [the docs about `combinator` module](https://docs.rs/winnow/latest/winnow/combinator/index.html) the entry for `dec_int` points to the page for  `dec_uint`. This small PR fixes that.

I'm not sure if it worth creating an issue for change so small, so I didn't create it 🙂


epage:
Thanks!

PR #466, opened 2024-02-08
docs: Update reference from `success` to `empty`

vwkd:
Remove reference to `success` deprecated in #440.

### DIFF
--- a/src/combinator/mod.rs
+++ b/src/combinator/mod.rs
@@ -82,7 +82,7 @@
 //! ## Remaining combinators
 //!
 //! - [`empty`]: Returns a value without consuming any input, always succeeds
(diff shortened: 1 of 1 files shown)


src/combinator/mod.rs
```diff
@@ -82,7 +82,7 @@
 //! ## Remaining combinators
 //!
 //! - [`empty`]: Returns a value without consuming any input, always succeeds
-//! - [`fail`]: Inversion of `success`. Always fails.
+//! - [`fail`]: Inversion of `empty`. Always fails.
```
epage, line 85:
```suggestion
//! - [`fail`]: Inversion of [`empty`]. Always fails.
```
Help catch problems like this in the future :)

epage:
Thanks!

PR #499, opened 2024-03-27
Remove useless escargot dependency

irevoire:
Hey I was comparing this project with nom and noticed you had an useless dependency in your `Cargo.toml`


epage:
Huh, looks like this was an over-aggressive copy/paste from another project in #176

PR #500, opened 2024-03-30
Add impl `ErrorConvert` for `ContextError`

Lexicality:
Quick fix for #498

I didn't touch TreeError because I couldn't see a particularly easy way of converting input types on it other than rewriting the entire tree which seemed a bit over the top.

I added the `InputError` -> `ContextError` conversion because I could rather than because I needed it, I can remove it if you don't want it.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -711,6 +711,13 @@ impl crate::lib::std::fmt::Display for ContextError<StrContext> {
     }
 }
 
+impl<C> ErrorConvert<ContextError<C>> for ContextError<C> {
+    #[inline]
+    fn convert(self) -> ContextError<C> {
+        self
+    }
+}
+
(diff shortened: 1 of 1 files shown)


Lexicality:
Huh, weird. It wouldn't build on my machine either but I assumed I had something configured wrong. :thinking:

epage:
> I added the InputError -> ContextError conversion because I could rather than because I needed it, I can remove it if you don't want it.

Let's remove that for now as its unrelated to the problem at hand and it would be good to understand the use case for why someone is doing it.

Lexicality:
> > I added the InputError -> ContextError conversion because I could rather than because I needed it, I can remove it if you don't want it.
> 
> Let's remove that for now as its unrelated to the problem at hand and it would be good to understand the use case for why someone is doing it.

done

Issue #406, opened 2023-12-27
Add `_temporary` variables to `seq!`


DJDuque:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### winnow version

0.5.31

### Describe your use case

Having `_temporary` variables that don't make it into the struct (but also don't get completely ignored like `_`) could be beneficial for cases like [this](https://github.com/MIDAS-rs/midasio/blob/cc2ed1e54f8fc40eb7bfabd79d490ec4e8ba38b2/src/event.rs#L115-L139).
The above `seq!` call could look something like:
```
 pub(crate) fn event_view<'a>(
    endian: Endianness,
) -> impl Parser<&'a [u8], EventView<'a>, ContextError> {
    seq! {EventView {
        event_id: u16(endian),
        trigger_mask: u16(endian),
        serial_number: u32(endian),
        timestamp: u32(endian),
        _event_size: u32(endian).verify(|&n| n >= 8),
        _banks_size: u32(endian).verify(&n| n == _event_size - 8),
        bank_views: dispatch! {u32(endian);
                1 => length_and_then(success(_banks_size), repeat_till0(padded_bank(bank_16_view(endian)), eof)),
                17 => length_and_then(success(_banks_size), repeat_till0(padded_bank(bank_32_view(endian)), eof)),
                49 => length_and_then(success(_banks_size), repeat_till0(padded_bank(bank_32a_view(endian)), eof)),
                _ => fail,
            }.map(|(bank_views, _)| bank_views),
    }}
}
```
which is easier to follow and understand.

This will also allow us to use this macro to do stuff similar to `nom-derive`'s anonimous fields mentioned [here](https://github.com/winnow-rs/winnow/issues/82#issuecomment-1513415419), e.g.
```
seq! {MyStruct {
    _b0: be_u8,
    li: success(_b0).map(|n| n >> 6),
    version: success(_b0).map(|n| (n >> 3) & 0b111),
    // ... etc
}}

### Describe the solution you'd like

I don't know how an implementation of this would look like

### Alternatives, if applicable

_No response_

### Additional Context

_No response_

epage:
[...]

PR #511, opened 2024-04-25
Resolve issue #406 for structures

DJDuque:
Resolve issue #406 for structures.

I gave a glance to the `seq!` implementation for tuples, but I am not sure I understand how `_` fields are elided. It is OK if you don't want to merge this until there is support for tuples as well (I could give it another go later, but I can't promise anything).


epage:
For anyone else following along, there is more discussion in #406.

PR #517, opened 2024-05-01
Update command in chapter 8

YouSafe:
While trying to run tests with the debug feature using the tutorial's command, I encountered an error:

```bash
$ cargo test --features debug
error: none of the selected packages contains these features: debug
```

After some investigation and learning how features in cargo work, I found that using `cargo test --features winnow/debug` resolved the issue. I recommend updating the tutorial to include this command to avoid future confusion for other users.


epage:
Thanks!

PR #520, opened 2024-05-04
chore: fixup parse_bin_digits in example

tisonkun:
TRIVIAL AS IS.

Please take a review @epage.


epage:
Thanks!  Found more cases with this so I went ahead and fixed all of them.

PR #525, opened 2024-05-21
feat: derive Default for (Stateful|Recoverable)

### LINKED ISSUE #524: derive `Default` for `Stateful`

### winnow version

0.6.8

### Describe your use case

I want to utilize the subparser pattern, to minimize the scope of the parser state.

Currently, my solution is like this:

```rust
fn with_state<S, State, O, ParseNext>(mut parse_next: ParseNext) -> impl Parser<S, O, ContextError>
where
    S: TokenStream,
    State: Default,
    ParseNext: Parser<Stateful<S, State>, O, ContextError>,
{
    move |input: &mut S| -> PResult<O> {
        let state = State::default();
        let input2 = std::mem::take(input);
        let mut stateful = Stateful {
            input: input2,
            state,
        };
        let output = parse_next.parse_next(&mut stateful);
        *input = stateful.input;
        output
    }
}
```

This works well, unless I want to use `Stateful<Stateful<S, State1>, State2>`, which doesn't implement `Default` even if both `S`, `State1` and `State2` implemented `Default`.

Derive `Default` for `Stateful`.

It's harmless for any other use case. Only when both `input` and `state` implemented `Default`, `Stateful<I, S>` will implemented `Default`.

_No response_

### Additional Context

_No response_

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

Resolve #524

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -176,6 +176,17 @@ where
 
 #[cfg(feature = "unstable-recover")]
 #[cfg(feature = "std")]
+impl<I, E> Default for Recoverable<I, E>
+where
+    I: Default + Stream,
+{
+    fn default() -> Self {
+        Self::new(I::default())
+    }
+}
+
+#[cfg(feature = "unstable-recover")]
+#[cfg(feature = "std")]
 impl<I, E> Recoverable<I, E>
 where
     I: Stream,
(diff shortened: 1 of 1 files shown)


src/stream/mod.rs
```diff
@@ -298,7 +298,7 @@ impl<I: Stream + crate::lib::std::fmt::Debug, E: crate::lib::std::fmt::Debug>
 /// let output = word.parse(input).unwrap();
 /// assert_eq!(state.get(), 1);
 /// ```
-#[derive(Clone, Copy, Eq, PartialEq)]
+#[derive(Clone, Copy, Default, Eq, PartialEq)]
```
epage, line 301: If we do this, we should probably do it to all of the Stream wrappers and to our custom Stream types, like `Bytes`.

epage: Ah, looked in the wrong file.  It looks like `Bytes` and `BStr` already `impl Default`.   Also, I had missed some explicit `impl Default`s.

It looks like only `Recoverable` is missing it.  Could you add it to that?

TennyZhuang: Thanks! I've added it.

TennyZhuang:
I forgot to add feature guard in yesterday's commit, and now it's fixed.

TennyZhuang:
Can the PR got merged? Thanks @epage

PR #533, opened 2024-06-03
fix: Remove AsBytes from some trait bounds

epage:
The `AsBytes` bounds came about because
- I prefer composing parsers where possible
- `Stream::Slice` is not guaranteed to be a `Stream` and making it so is messy

This change came about because of winnow-rs/winnow#528


ckoehler:
This enables all the binary integer parsers for us, thanks! Been testing it in different combos with `length_and_then` and `length_take`, and it seems to work as expected! Appreciate the work!

PR #550, opened 2024-06-24
docs(tut): Add cross ref to chapter 6

VorpalBlade:
This helps when users try to use the docs as a reference to solve their issue at hand (as opposed to reading in order and remembering everything).

Related to discussion #544

### DIFF
--- a/src/_tutorial/chapter_7.rs
+++ b/src/_tutorial/chapter_7.rs
@@ -567,6 +567,9 @@
 //! - Correctly reporting columns with unicode
 //! - Conforming to a specific layout
 //!
+//! As a reminder [`chapter_6`] demonstrated how to convert a [`PResult`] to a standard
+//! Rust [`Result`], and this section assumes that has been done.
+//!
 //! For example, to get rustc-like errors with [`annotate-snippets`](https://crates.io/crates/annotate-snippets):
 //! ```rust
 //! # use winnow::prelude::*;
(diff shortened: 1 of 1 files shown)


VorpalBlade:
I don't think that "coveralls" report is correct. I haven't changed any of the files it claims have new missed lines...

src/_tutorial/chapter_7.rs
```diff
@@ -567,6 +567,9 @@
 //! - Correctly reporting columns with unicode
 //! - Conforming to a specific layout
 //!
+//! As a reminder [`chapter_6`] demonstrated how to convert a [`PResult`] to a standard
+//! Rust [`Result`], and this section assumes that has been done.
+//!
```
epage, line 572: I feel this doesn't quite serve the intended role.
- Everything up to this point was already relying on this in this chapter
- If we had this, I feel like it should call out the mechanism whereby this is done or else it feels like its pointing back to nothing
- I believe our callbacks like this are usually done when we reference much earlier chapters.  What some chapters do for this type of thing is, at the very type, have a transition statement.  For example, we could have something like "With `Parser::parse` converting `PResult` into `Result` for use in our application, let's see how we can polish the quality of the reported error message".  Note that that example statement is meant only to serve as an example and I created it without much deep thought.

epage:
To be clear, I gave a first-order approximation of how this could be reworked

>  For example, we could have something like
> "With Parser::parse converting PResult into Result for use in our application, let's see how we can polish the quality of the reported error message".

PR #553, opened 2024-06-29
chore: Correct ^ error position in unicode strings

gzigzigzeo:
# Problem

Given we have the following string:

```rust
let input = r#""Hello, 世界! 🌍\*""
```

Let's say that `\*` represents an invalid sequence. By default `^` error symbol would be misplaced in the standard output:

```sh
  |
7 |     "Hello, 世界! 🌍\*"
  |                ^
```

This happens because symbols like `世界🌍` have double width in fixed width fonts.

The more emojis precede an error, the more the error pointer shifts. It is confusing, especially if errors are displayed to the end user.

# Solution

Adds `unicode-width` feature, which calculates correct offset using `unicode-width` crate.

```sh
  |
7 |     "Hello, 世界! 🌍\*"
  |                    ^
```

Thank you for this great crate!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1571,6 +1571,8 @@ dependencies = [
  "snapbox",
  "term-transcript",
  "terminal_size",
+ "unicode-width",
+ "winnow",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


epage:
We intentionally keep the error simple as these types of questions are dependent on where you are rendering to and how they handle different unicode characters, as well as keeping the dep tree small.

If you'd like to discuss this further, feel free to create an issue.  In general, I recommend issues for non-trivial changes for discussing the solution space and encourage PRs only for reviewing an implementation.

gzigzigzeo:
Thanks, will open an issue then.

PR #554, opened 2024-07-01
doc: Add a blank line between paragraph and image

joshtriplett:
This prevents the large image from appearing inline in the last line of
the paragraph.

Without this fix:

image


epage:
Thanks!  Must have been a regression from when we moved it to the tutorial as I had looked at t he previous page many times and it didn't look like that.

PR #599, opened 2024-09-29
Adjust `Repeat::fold` doc parameter name to that of implementation

SRv6d:
Adjust `Repeat::fold` doc parameter name to match that of implementation.


epage:
Thanks!

PR #541, opened 2024-06-14
RFC: Add `recognize_till`

9999years:
This is a variant of `repeat_till` that recognizes the repeated portion of the input instead of accumulating it.

See: [“Recognizing `repeat_till0` to `()` without the suffix”](https://github.com/winnow-rs/winnow/discussions/336#discussioncomment-7105907).

I have found this combinator useful in my own code for recognizing ambiguous delimiters.

This code needs to be cleaned up before being merged (we'd probably want a `Range` argument, for example), so consider this a proof of concept. If you're on board with the idea, let me know what the next steps are. Otherwise, feel free to close this PR. Thanks!

PR #604, opened 2024-10-14
Add `ParseError::with_input`

9999years:
This lets you replace the input field in a `ParseError`, e.g. to transform a `ParseError<&str, E>` into a `ParseError<String, E>`. I'd like this so that I can have an owned `ParseError` type, but only if a top-level parse fails.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1164,7 +1164,8 @@ pub struct ParseError<I, E> {
 }
 
 impl<I: Stream, E: ParserError<I>> ParseError<I, E> {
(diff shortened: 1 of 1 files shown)


epage:
We [ask](https://github.com/winnow-rs/winnow/blob/main/CONTRIBUTING.md#pull-requests) that new features start as issues to explore the underlying use case and requirements and then discuss how we want to solve the problem.  We then move onto PRs to discuss the implementation.

9999years:
Done: #606

I now see that you also told me to open an issue first (quite kindly) on #541 — apologies for making the same mistake twice. I assumed the change was small enough to be uncontroversial, but I can respect having all changes go through the same process.

Issue #131, opened 2023-02-07
Pratt parsing support


epage:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### winnow version

0.2.0

### Describe your use case

This would be a performant way of handling precedence in parsing

See zesterer/chumsky#51

### Describe the solution you'd like

See zesterer/chumsky#144

epage:
For more technical details, see
- https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
- https://matklad.github.io/2020/04/15/from-pratt-to-dijkstra.html

Zij-IT:
Heyyo! I was the one to handle a significant portion of the Pratt parser implementation that is available in chumsky :)

I would have to familiarize myself with how y'all do things here at Winnow, but I would be happy to take a crack at getting a Pratt parsing implementation going for Winnow.

Do you have an idea already for an API that you would prefer, or guidelines on how to keep the API winnow-style :D I'd love to contribute to getting some cross-pollination going between two great parsing crates!

epage:
That would be greatly appreciated!

I have no formal background in parsing and only know of Pratt Parsing from matklad's posts and then your work with chumksy.  I'm not even sure where or if I would use it in my own work.  So its hard for me to give too specific of guidance on the API.

Overall, the philosophical divergence between chumsky and winnow is that chumsky is a framrwork while winnow is a toolbox.  Chumsky owns the process and users plug into that (e.g. users are expected to *not* implement the `Parser` trait).  Winnow provides common primitives and allows users to build up their parser out of that, as needed.  We freely encourage mixing of imperative and functional / declarative parsers.  Unsure how that might apply but thought I'd point that out.

One challenge I repeatedly run into is that when we give too high level of a helper, it also constrains the user.  A trivial example of this is `Parser::try_map` which doesn't let you cut-to-top the error from `try_map` separate from the error from the parser it wraps (#180).

epage:
> Its strength (how tightly it binds. e.g. + binds weaker than *)

So this is "precedence", right?

> `Right(1),`

I assume this is associativity and strength combined.  Why have them together?

> pratt could also be free-standing function which accepts two parsers: 

[...]

PR #614, opened 2024-11-12
feat: implement Pratt parsing

39555:
Closes: #131 

This is the initial implementation of the `operator precedence` parser based on the Pratt algorithm https://en.wikipedia.org/wiki/Operator-precedence_parser#Pratt_parsing. 

The interface:
```rust
precedence(
    digit1.try_map(|d: &str| d.parse::<i32>()),
    (
        "-".value(2).prefix(|x| -1 * x),
        "+".value(2).prefix(|x| x),
        "!".value(2).postfix(|x| factorial(x)),
        "+".value(0).infix(|a, b| a + b),
        "-".value(0).infix(|a, b| a + b),
        "*".value(1).infix(|a, b| a * b),
        "/".value(1).infix(|a, b| a / b),
    ),
)
.parse_next(i)
```
Any parser that returns a number can be considered as operator with its `binding power`. To convert the parser into `prefix`, `infix` or `postfix`, we use the extension trait on top of the `impl Parser`.

### Binding power, associativity, Pratt and Dijkstra

The Pratt algorithm only uses the `binding power`. `Left` and `right` associativity are used in Dijkstra's [`Shunting Yard`](https://en.wikipedia.org/wiki/Shunting_yard_algorithm) algorithm.  (It seems like`chumsky` incorrectly names its implementation as `pratt` while using `shunting yard` internally).

I named the combinator `precedence` with aliases to `pratt`, `shunting yard`, `precedence climbing`, `separated`. If any of the names are confusing, we may consider better alternatives.

The implementation is partially inspired by https://gist.github.com/ilonachan/3d92577265846e5327a3011f7aa30770.
But it doesn't use any `Rc<...>` allocations for closures; instead, it uses uses `& RefCell<& dyn ...>` references.

Operators from the tuple are statically dispatched by the requested affix so `prefix.as_postfix()` will return `fail` while `prefix.as_prefix()` with return the actual parser. This may have a performance issue because it iterates through the whole tuple each time the parser is requested for any affix:

<details><summary>See the trace:</summary><p>

```
> postfix                                                                                      | "+4"∅
 > opt                                                                                         | "+4"∅
  > alt                                                                                        | "+4"∅
   > fail                                                                                      | "+4"∅
   < fail                                                                                      | backtrack
   > fail                                                                                      | "+4"∅
   < fail                                                                                      | backtrack
   > "!"                                                                                       | "+4"∅
   < "!"                                                                                       | backtrack
   > fail                                                                                      | "+4"∅
   < fail                                                                                      | backtrack
   > fail                                                                                      | "+4"∅
   < fail                                                                                      | backtrack
   > fail                                                                                      | "+4"∅
   < fail                                                                                      | backtrack
   > fail                                                                                      | "+4"∅
   < fail                                                                                      | backtrack
  < alt                                                                                        | backtrack
 < opt                                                                                         | +0
< postfix
```

</p></details>

(I tried to make a some sort of a linked list where each parser holds an index to the next one with the same affix but, it quickly became a nightmare of lifetimes, complexity, and compilation errors so it was rejected).

### Performance

A quick criterion benchmark with a simple input  &str `1-2*4+12-561-5-6*6/9-3+1*-2*4-758*3` shows:
```
winnow   time: [191.25 ns 191.56 ns 191.94 ns]
chumsky time: [335.63 ns 335.93 ns 336.42 ns]
```

### TODO

- [ ] Add more tests from here https://gist.github.com/ilonachan/3d92577265846e5327a3011f7aa30770 and here https://github.com/rust-bakery/nom/pull/1362/files
- [ ] Implement a full example with ternary and other fun operators
- [ ] Performance benchmarks for the example
- [ ] Documentation with examples
- [ ] Implement the api for Vec<..> and &[...]
- [ ] Streaming support
- [ ] Hide behind a feature flag`unstable-pratt`

### DIFF
--- a/src/combinator/mod.rs
+++ b/src/combinator/mod.rs
@@ -164,6 +164,7 @@ mod core;
 mod debug;
 mod multi;
 mod parser;
+mod precedence;
 mod sequence;
 
 #[cfg(test)]
@@ -174,6 +175,7 @@ pub use self::core::*;
 pub use self::debug::*;
 pub use self::multi::*;
 pub use self::parser::*;
+pub use self::precedence::*;
 pub use self::sequence::*;
 
 #[allow(unused_imports)]
--- a/src/combinator/precedence.rs
+++ b/src/combinator/precedence.rs
@@ -1,0 +1,481 @@
+use core::cell::RefCell;
+
+use crate::{
+    combinator::{alt, fail, opt, trace},
+    error::{ErrMode, ParserError},
+    stream::{Stream, StreamIsPartial},
+    PResult, Parser,
+};
+
+/// An adapter for the [`Parser`] trait to enable its use in the [`precedence`] parser.
+pub trait PrecedenceParserExt<I, E> {
+    /// Specifies that the parser is a `unary` `prefix` operator within a [`precedence`] parser.
+    ///
+    /// In most languages, operators such negation: `-`, `not` or `!`, dereferencing: `*`, etc. are prefix unary operators.
+    ///
+    /// The argument `fold` is a fold function that defines how to combine the operator and operand into a new expression.
+    /// It must have the following signature:
+    /// ```ignore
+    /// impl Fn(O) -> O
+    /// ```
+    #[inline(always)]
+    fn prefix<F, O>(self, fold: F) -> Prefix<Operator<F, Self>>
+    where
+        F: UnaryOp<O>,
+        Self: Sized,
+    {
+        Prefix(Operator::new(self, fold))
+    }
+    /// Specifies that the parser is a `unary` `postfix` operator within a [`precedence`] parser.
+    ///
+    /// Operators like the factorial `!` are postfix unary operators.
+    ///
+    /// The argument `fold` is a fold function that defines how to combine the operator and operand into a new
+    /// expression. It must have the following signature:
+    /// ```ignore
+    /// impl Fn(O) -> O
+    /// ```
+    #[inline(always)]
+    fn postfix<F, O>(self, fold: F) -> Postfix<Operator<F, Self>>
+    where
+        F: UnaryOp<O>,
+        Self: Sized,
+    {
+        Postfix(Operator::new(self, fold))
+    }
+    /// Specifies that the parser is a `binary` `infix` operator within a [`precedence`] parser.
+    ///
+    /// Operators like factorial `+`, `-`, `*`, `/` are infix binary operators.
+    ///
+    /// The argument is a fold function that defines how to combine the operator and two operands into a new
+    /// expression. It must have the following signature:
+    /// ```ignore
+    /// impl Fn(O, O) -> O
+    /// ```
+    #[inline(always)]
+    fn infix<F, O>(self, fold: F) -> Infix<Operator<F, Self>>
+    where
+        F: BinaryOp<O>,
+        Self: Sized,
+    {
+        Infix(Operator::new(self, fold))
+    }
+}
+impl<I, E, T: Parser<I, usize, E>> PrecedenceParserExt<I, E> for T where I: Stream {}
+
+/// `NewType` that indicates this type is a prefix operator a [`precedence`] parser.
+/// See: [`PrecedenceParserExt::prefix`]
+///
+/// Can hold and arbitrary parser, such as a tuple of multiple operator parsers: `(Operator<...>, Operator<...>)`
+pub struct Prefix<T>(T);
+
+/// `NewType` that indicates this type is a postfix operator a [`precedence`] parser.
+/// See: [`PrecedenceParserExt::postfix`]
+pub struct Postfix<T>(T);
+
+/// `NewType` that indicates this type is a infix operator within a [`precedence`] parser.
+/// See: [`PrecedenceParserExt::infix`]
+pub struct Infix<T>(T);
+
+/// Implementation of the operator parser for the [`precedence`] parser.
+pub struct Operator<OperatorFunc, OperatorParser> {
+    // We use two different `ReffCell`s to enable mutable borrowing within the recursion
+    // while holding a reference to the predicate `op`:
+    // ```
+    //      let lhs = ...;
+    //      let op: &ReffCell<dyn BinaryOp<...>> = operator.parse_next(i); // calls `operator.parser.borrow_mut().parse_next(i)`
+    //      let rhs = recursion(&operator);
+    //      let result = op.borrow_mut().fold_binary(lhs, rhs);
+    // ```
+    op: RefCell<OperatorFunc>,
+    parser: RefCell<OperatorParser>,
+}
+
+impl<OperatorFunc, OperatorParser> Operator<OperatorFunc, OperatorParser> {
+    /// Creates a new [`Operator`] from a parser and a predicate
+    #[inline(always)]
+    pub fn new(parser: OperatorParser, predicate: OperatorFunc) -> Self {
+        Self {
+            op: RefCell::new(predicate),
+            parser: RefCell::new(parser),
+        }
+    }
+}
+
+/// Type-erased unary predicate that folds an expression into a new expression.
+/// Useful for supporting not only closures but also arbitrary types as operator predicates within the [`precedence`] parser.
+pub trait UnaryOp<O> {
+    /// Invokes the [`UnaryOp`] predicate.
+    fn fold_unary(&mut self, o: O) -> O;
+}
+/// Type-erased binary predicate that folds two expressions into a new expression similar to
+/// [`UnaryOp`] within the [`precedence`] parser.
+pub trait BinaryOp<O> {
+    /// Invokes the [`BinaryOp`] predicate.
+    fn fold_binary(&mut self, lhs: O, rhs: O) -> O;
+}
+
+impl<O, F> UnaryOp<O> for F
+where
+    F: Fn(O) -> O,
+{
+    #[inline(always)]
+    fn fold_unary(&mut self, o: O) -> O {
+        (self)(o)
+    }
+}
+impl<O, F> BinaryOp<O> for F
+where
+    F: Fn(O, O) -> O,
+{
+    #[inline(always)]
+    fn fold_binary(&mut self, lhs: O, rhs: O) -> O {
+        (self)(lhs, rhs)
+    }
+}
+
+impl<'s, UO, O, I, P, E> Parser<I, (&'s RefCell<dyn UnaryOp<O>>, usize), E> for &'s Operator<UO, P>
+where
+    UO: UnaryOp<O> + 'static,
+    I: Stream + StreamIsPartial,
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn parse_next(
+        &mut self,
+        input: &mut I,
+    ) -> PResult<(&'s RefCell<dyn UnaryOp<O> + 'static>, usize), E> {
+        let power = self.parser.borrow_mut().parse_next(input)?;
+        Ok((&self.op, power))
+    }
+}
+impl<'s, BO, O, I, P, E> Parser<I, (&'s RefCell<dyn BinaryOp<O>>, usize), E> for &'s Operator<BO, P>
+where
+    BO: BinaryOp<O> + 'static,
+    I: Stream + StreamIsPartial,
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn parse_next(
+        &mut self,
+        input: &mut I,
+    ) -> PResult<(&'s RefCell<dyn BinaryOp<O> + 'static>, usize), E> {
+        let power = self.parser.borrow_mut().parse_next(input)?;
+        Ok((&self.op, power))
+    }
+}
+
+/// Ability to request a parser of the specified affix from the [`impl Parser`](Parser) object.
+pub trait AsPrecedence<I: Stream, Operand: 'static, E: ParserError<I>> {
+    /// Interprets a parser as a [`PrecedenceParserExt::prefix`] parser that returns an `unary
+    /// predicate` [`UnaryOp`] and a `binding power` as its parsing result.
+    #[inline(always)]
+    fn as_prefix(&self) -> impl Parser<I, (&RefCell<dyn UnaryOp<Operand>>, usize), E> {
+        fail
+    }
+    /// Interprets a parser as a [`PrecedenceParserExt::postfix`] parser that returns an `unary
+    /// predicate` [`UnaryOp`] and a `binding power` as its parsing result.
+    #[inline(always)]
+    fn as_postfix(&self) -> impl Parser<I, (&RefCell<dyn UnaryOp<Operand>>, usize), E> {
+        fail
+    }
+    /// Interprets a parser as a [`PrecedenceParserExt::infix`] parser that returns a `binary
+    /// predicate` [`BinaryOp`] and a `binding power` as its parsing result.
+    #[inline(always)]
+    fn as_infix(&self) -> impl Parser<I, (&RefCell<dyn BinaryOp<Operand>>, usize), E> {
+        fail
+    }
+}
+
+impl<'s, F, O, I, P, E> Parser<I, (&'s RefCell<dyn UnaryOp<O>>, usize), E>
+    for &'s Prefix<Operator<F, P>>
+where
+    F: UnaryOp<O> + 'static,
+    I: Stream,
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn parse_next(
+        &mut self,
+        input: &mut I,
+    ) -> PResult<(&'s RefCell<dyn UnaryOp<O> + 'static>, usize), E> {
+        let power = self.0.parser.borrow_mut().parse_next(input)?;
+        Ok((&self.0.op, power))
+    }
+}
+
+impl<F, O: 'static, I, P, E> AsPrecedence<I, O, E> for Prefix<Operator<F, P>>
+where
+    F: UnaryOp<O> + 'static,
+    I: Stream + StreamIsPartial,
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn as_prefix(&self) -> impl Parser<I, (&RefCell<dyn UnaryOp<O>>, usize), E> {
+        &self.0
+    }
+}
+
+impl<F, O: 'static, I, P, E> AsPrecedence<I, O, E> for Postfix<Operator<F, P>>
+where
+    F: UnaryOp<O> + 'static,
+    I: Stream + StreamIsPartial,
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn as_postfix(&self) -> impl Parser<I, (&RefCell<dyn UnaryOp<O>>, usize), E> {
+        &self.0
+    }
+}
+impl<F, O: 'static, I, P, E> AsPrecedence<I, O, E> for Infix<Operator<F, P>>
+where
+    F: BinaryOp<O> + 'static,
+    I: Stream + StreamIsPartial,
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn as_infix(&self) -> impl Parser<I, (&RefCell<dyn BinaryOp<O>>, usize), E> {
+        &self.0
+    }
+}
+
+macro_rules! impl_parser_for_tuple {
+    () => {};
+    ($head:ident $($X:ident)*) => {
+        impl_parser_for_tuple!($($X)*);
+        impl_parser_for_tuple!(~ $head $($X)*);
+    };
+    (~ $($X:ident)+) => {
+
+        #[allow(unused_variables, non_snake_case)]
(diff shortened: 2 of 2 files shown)


src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+    /// Interprets a parser as a [`PrecedenceParserExt::infix`] parser that returns a `binary
+    /// predicate` [`BinaryOp`] and a `binding power` as its parsing result.
+    #[inline(always)]
+    fn as_infix(&self) -> impl Parser<I, (&RefCell<dyn BinaryOp<Operand>>, usize), E> {
+        fail
+    }
+}
+
+impl<'s, F, O, I, P, E> Parser<I, (&'s RefCell<dyn UnaryOp<O>>, usize), E>
```
epage, line 192: Should this be moved up next to `UnaryOp`, rather than splitting `AsPrecedence` trait with the impls?

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+
+macro_rules! impl_parser_for_tuple {
+    () => {};
+    ($head:ident $($X:ident)*) => {
+        impl_parser_for_tuple!($($X)*);
+        impl_parser_for_tuple!(~ $head $($X)*);
+    };
+    (~ $($X:ident)+) => {
+
+        #[allow(unused_variables, non_snake_case)]
```
epage, line 256: Why allow `non_snake_case`?

epage, line 256: Whats getting unused?

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+where
+    F: Fn(O) -> O,
+{
+    #[inline(always)]
+    fn fold_unary(&mut self, o: O) -> O {
+        (self)(o)
+    }
+}
+impl<O, F> BinaryOp<O> for F
+where
```
epage, line 128: I tend to prefer that impls be closely organized with their trait

src/combinator/mod.rs
```diff
@@ -174,6 +175,7 @@ pub use self::core::*;
 pub use self::debug::*;
 pub use self::multi::*;
 pub use self::parser::*;
+pub use self::precedence::*;
```
epage, line 178: We are dumping a lot of stray types into `combinator`.  The single-line summaries should make it very easy to tell they are related to `precedence` (maybe be the first word) and somehow help the user know what, if any, they should care about

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn as_infix(&self) -> impl Parser<I, (&RefCell<dyn BinaryOp<O>>, usize), E> {
+        &self.0
+    }
+}
+
+macro_rules! impl_parser_for_tuple {
```
epage, line 248: So we find the right "mode" by falling through an `alt` of parsers for that mode and then get the weight and fold function from the first success.

Looking at the example
```rust
precedence(
    digit1.try_map(|d: &str| d.parse::<i32>()),
    (
        "-".value(2).prefix(|x| -1 * x),
        "+".value(2).prefix(|x| x),
        "!".value(2).postfix(|x| factorial(x)),
        "+".value(0).infix(|a, b| a + b),
        "-".value(0).infix(|a, b| a + b),
        "*".value(1).infix(|a, b| a * b),
        "/".value(1).infix(|a, b| a / b),
    ),
)
.parse_next(i)
```
the alt seems unfortunate for performance reasons.  `dispatch` can be used in  this case and I suspect a lot of other cases.  It would be a big help to find a way to set this up so `dispatch` can be used as well, e.g.
```rust
precedence(
    digit1.try_map(|d: &str| d.parse::<i32>()),
    dispatch!{any;
        '-' => prefix(2, |x| -1 * x),
        '+' => prefix(2, |x| x),
        '!' => postfix(2, |x| factorial(x)),
        '+' => infix(0, |a, b| a + b),
        '-' => infix(0, |a, b| a + b),
        '*' => infix(1, |a, b| a * b),
        '/' => infix(1, |a, b| a / b),
        _ => fail,
    ),
)
.parse_next(i)
```
(note: I might give other suggestions that run counter to this, we'll need to weigh out each one and figure out how we want to balance the different interests)

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+    P: Parser<I, usize, E>,
+    E: ParserError<I>,
+{
+    #[inline(always)]
+    fn parse_next(
+        &mut self,
+        input: &mut I,
+    ) -> PResult<(&'s RefCell<dyn UnaryOp<O> + 'static>, usize), E> {
+        let power = self.parser.borrow_mut().parse_next(input)?;
+        Ok((&self.op, power))
```
epage, line 150: The parser is returning the power.

Its been a while since I've dealt with precedence.  Is having multiple items with the same power a requirement?  I'm wondering if we could assume the power is the index into the tuple/array

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+/// This algorithm is similar to the [Shunting Yard](https://en.wikipedia.org/wiki/Shunting_yard_algorithm) algorithm
+/// in that both are linear, both use precedence and binding power, and both serve the same purpose.
+/// However, the `Shunting Yard` algorithm additionally uses `left` and `right` associativity,
+/// while `Pratt` parsing only relies on binding power.
+#[doc(alias = "pratt")]
+#[doc(alias = "separated")]
+#[doc(alias = "shunting yard")]
+#[doc(alias = "precedence climbing")]
+#[inline(always)]
+pub fn precedence<I, ParseOperand, Operators, Operand: 'static, E>(
```
epage, line 347: Organizationally, O prefer the "top level" thing going first and then branching out from there.  In this case, `precedence` is core.

epage (commented):
Wow, this looks great.  You put a lot of work into this and I see care is taken in cases like the handling of infinite loops and such

src/combinator/mod.rs
epage, line 1: >  Streaming support

This looks to be agnostic of streaming support like `separated` is

epage, line 1: > Hide behind a feature flagunstable-pratt

If this is going to start off unstable, then its fine noting most of my feedback in the "tracking" issue and not resolving all of it here

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+
+use crate::{
+    combinator::{alt, fail, opt, trace},
+    error::{ErrMode, ParserError},
+    stream::{Stream, StreamIsPartial},
+    PResult, Parser,
+};
+
+/// An adapter for the [`Parser`] trait to enable its use in the [`precedence`] parser.
+pub trait PrecedenceParserExt<I, E> {
```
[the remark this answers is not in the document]
39555: Thanks for reviewing! Do you suggest making `prefix`, `infix`..  standalone functions?  I followed the concern in the attached issue to make `<parser>.prefix(...)` work:
> So I assume we'd go with
> ```
> let calc = pratt(
>     digits1.map(Expr::Int),
>     (
>         '-'.prefix(Right(1), |r| unary(r, Op::Neg));
>         '+'.infix(Left(0), |l, r| binary(l, r, Op::Add));
>         '!'.prefix(Right(3), |r| unary(r, Op::Fact));
>     )
> );
> ```
These `prefix`, `infix`, `postfix` are strange beasts in this implementation. They are both grammar-level constructs and value processing functions because the data processing is required by the parser to advance further.

I though about the interface like `prefix(0, "-").map(|(a, b)| a - b)`. It looks clean but what should be the default `operator` when we don't call `map`. For a `unary` it is easy -- just return the argument, but for a `binary` it is unclear how to merge operands.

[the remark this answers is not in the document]
epage: Ha!  I previously looked at this and viewed the associativity as not grammar level and now looked at it and viewed it as grammar level.  Normally trait methods are reserved for purely operating on the data, like `map`.  This is actually changing how we parse the results.  The one weird middle ground we have is `cut_err`.  Unsure if this should be treated like that or not.

I also feel like a bunch of free functions floating around might not be the easiest for discovery and use.

I also suspect the ultimate answer will be dependent on the answer to other questions as I alluded to the interplay of my comments in https://github.com/winnow-rs/winnow/pull/614#discussion_r1838614798

src/combinator/precedence.rs
39555: Thanks for catching that. This is required only for the second trait in the macro because of `let ($($X)*) = self;`

src/combinator/precedence.rs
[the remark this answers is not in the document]
39555: Maybe `expression`? Similar to how `combine` names this parser https://docs.rs/combine-language/latest/combine_language/fn.expression_parser.html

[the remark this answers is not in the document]
epage: Oh interesting, I wasn't aware of that crate.  We should at least add `expression_parser` to our aliases.

We likely should pick one and deal with it and rename it if a better name comes up.  `expression` seems just as good as any other.  It fits with how this would be used in a language grammar.

src/combinator/precedence.rs
epage: In other macros, we define separate terms for trait bounds, variables, etc.  Is it worth doing that here?

src/combinator/precedence.rs
39555: Another idea. The implementation in `nom` https://github.com/rust-bakery/nom/pull/1362 uses 3 parsers one for each affix in the function signature. It has an advantage that you can use `dispatch` and there is no `prefix`, `postfix`.. functions. The downside is that from just the function signature it's hard to remember which parser is which. And also `fail` is used if the parser is not provided.

Maybe a`builder` pattern would be ideal e.g:
```rust
precedence(digit1.parse_to::<i32>())
    .prefix(dispatch!{any; 
         "+" => empty.map(|_| move |a, b| a + b)
         "-" => empty.map(|_| move |a, b| a + b)
   })
   .postfix(dispatch!{any; ...})
   .infix(dispatch!{any; ...})
   .build()
.parse_next(i)
```
Methods are similar to `repeat().fold`. This approach would eliminate stray functions, reduce the performance impact of iterating over each parser for each affix, and allow users to use dispatch or any custom logic they want.

I’m not sure yet how this would interact with binding power. I'm going to research whether we can drop the explicit index argument and rely on array indexing, how binding power interact between affixes .etc.

39555: I came across another idea. Instead of dispathing by the parser type maybe we could dispatch by the input type:
```
enum Input<'i, I>{
   Prefix(&'i mut I),
   Postfix(&'i mut I),
   Infix(&'i mut I)
}
```
then it would be possible to (I'm not sure yet if Deref would work or not)
```rust 
dispatch!{any;
        '-' => prefix(2, |x| -1 * x),
        '+' => prefix(2, |x| x),
        '!' => postfix(2, |x| factorial(x)),
        '+' => infix(0, |a, b| a + b),
        '-' => infix(0, |a, b| a + b),
        '*' => infix(1, |a, b| a * b),
        '/' => infix(1, |a, b| a / b),
        _ => fail,
    )
```
then `prefix` parsers would do
```rust
if let Prefix() = input {
   ...
} else {
  fail
}
```

src/combinator/precedence.rs
```diff
@@ -0,0 +1,481 @@
...
+    E: ParserError<I>,
+{
+    trace("precedence", move |i: &mut I| {
+        let result = precedence_impl(i, &mut parse_operand, &ops, 0);
+        result
+    })
+}
+
+// recursive function
+fn precedence_impl<I, ParseOperand, Operators, Operand: 'static, E>(
```
epage, line 364: Is there a way we can do this without recursion?  Parser authors are generally cautious around recursion due to blowing up the stack.

Hmm, seems all of the other ones use recursion

src/combinator/precedence.rs
epage, line 347: > Binding power, associativity, Pratt and Dijkstra

The Pratt algorithm only uses the binding power. Left and right associativity are used in Dijkstra's [Shunting Yard](https://en.wikipedia.org/wiki/Shunting_yard_algorithm) algorithm. (It seems likechumsky incorrectly names its implementation as pratt while using shunting yard internally).

As someone not heavily versed in this stuff, what would be the trade off of pratt vs shunting yard?  It seems chumsky, combine-language, and the proposed nom pratt parser all have associativity.

src/combinator/precedence.rs
epage: This is nice because we only run the needed parsers (less `alt`, no `dispatch` just to get an error).

I agree about the opacity of parameter order.  Its also not great if a parameter is unused (e.g. no postfix).

If we go with the "builder" approach, we are again violating the API design guidelines as discussed in https://github.com/winnow-rs/winnow/pull/614#discussion_r1837390666.  However, it does make it nice for naming the parameters and only requiring what is needed.

Maybe one way of looking at this is the error reporting.  What happens if I use an operator in the wrong location?

epage: Not quite getting the input type and dealing with custom input types in the middle of a parser adds a lot of complexity.

src/combinator/precedence.rs
39555: While looking at it I found a potential error. My current implementation is unsound. Consider the `2 + 4 + 7`. In the current parser it evaluates as `(4 + 7) + 2` instead of `(2 + 4) + 7` when operators have equal binding power . That is why the algorithm uses associativity. While `+` or `*` are `commutative` someone might want to parse function calls for example

src/combinator/precedence.rs
39555: I was wrong about `left` and `right`. I updated the pr description. We also need that `left` and `right`. About shunting yard: I read that https://stackoverflow.com/questions/13634393/how-do-pratt-parsers-compare-to-other-parser-types-and-why-are-they-used-so-litt and https://news.ycombinator.com/item?id=19201396. And this cool repo https://github.com/bourguet/operator_precedence_parsing. The difference is mainly a recursion.

src/combinator/precedence.rs
39555: Same with the power based on the index. the `("+", "-")` would evaluate `2 + 4 - 7` as `2 + (4 - 7)` but the `("-", "+")` would be `(2 + 4) - 7`. It might be important in some cases (I didn't find one yet). One solution is to group operators with the same power but it hurts `dispatch`:
```
precedence( (
("+", "-"),
"*"
))
```

src/combinator/precedence.rs
epage, line 364: From https://github.com/winnow-rs/winnow/pull/614#discussion_r1839893540

> While looking at it I found a potential error. My current implementation is unsound. Consider the 2 + 4 + 7. In the current parser it evaluates as (4 + 7) + 2 instead of (2 + 4) + 7 when operators have equal binding power . That is why the algorithm uses associativity. While + or * are commutative someone might want to parse function calls for example

src/combinator/precedence.rs
epage: Thanks for clarifying!

src/combinator/precedence.rs
epage: So the relevant parts of this will be handled through
- https://github.com/winnow-rs/winnow/pull/614#discussion_r1838835199
- https://github.com/winnow-rs/winnow/pull/614#discussion_r1841200693

39555:
Another difference from the #618 algorithm is that this recursive one backtracks when the left binding power of an operand is less than the current right binding power. It could affect the performance if parsing the operators is expensive.
```
 1     +     2     +     3
    1     2     1     2
          -     -
-> "1 + 2 +"   ~^ 1 < 2
-> "1 + 2"
-> "3"             +
```
The #618 always stacks all parsed operators into a vector.

39555:
Operator callbacks should be able to return `impl Parser` or receive the input in order to provide ability to kick additional parsing. For example to parse an index operator, a function call, or a ternary operator. How this is done here https://gist.github.com/ilonachan/3d92577265846e5327a3011f7aa30770#file-nom-pratt-parser-rs-L198

```rust
fn ternary(cond: Expr) -> ... {
  move |i: &mut Input| {
        let (a, b) = separated_pair(pratt_parser , ':' , pratt_parser).parse_next(i)?;
        Ok(Expr::Ternary(cond, a, b)
  }
}
fn pratt_parser() -> ... {
    precedence(
        ...,
        trace("postfix", '?'.value((5, ternary))),
        ...
    )
}

```

src/combinator/precedence.rs
```diff
@@ -0,0 +1,193 @@
...
+                            '(' => delimited('(',  parser(), ')'),
+                            _ => digit1.parse_to::<i32>()
+                        },
+                        space0,
+                    ),
+                ),
+                trace(
+                    "prefix",
+                    dispatch! {any;
+                        '+' => empty.value((9, (|_: &mut _, a| Ok(a)) as _)),
```
epage, line 151: I forgot to ask, why is `as _` needed everywhere?

39555: The compiler don't want to cast `{closure34234234}` into function pointer `fn(&mut I, O) -> O` automatically:(

epage: Why do we need it to be a function pointer rather than a `dyn` or an `impl`?

I assume https://github.com/winnow-rs/winnow/pull/618#discussion_r1842381182 is related.

epage: If we provided an `infix` function, the compiler would do the coercion for us, right?

39555: I just checked. Yes this simple guidance works
```rust
fn infix<I, O>(
    a: Assoc,
    f: fn(&mut I, O, O) -> PResult<O>,
) -> (
    Assoc,
    fn(&mut I, O, O) -> PResult<O>,
) {
    (a, f)
}
...

"*".value(infix(Assoc::Right(28), |_: &mut _, a, b| Ok(Expr::Pow(Box::new(a), Box::new(b)))))
```

39555:
While trying to update the benchmark I used `bumpalo` and encountered a limitation of function pointers. I can't capture the arena allocator to create a Box:

```rust
|i: &mut &'i str, a| Ok(Expr::PreIncr(Box::new_in(a, bump)))
```

And it seems it is not possible to use `&dyn Fn(&mut I)`. At least, I can't:
```rust
in type `&'f (dyn Fn(&'i mut I, Operand) -> Result<Operand, ErrMode<E>> + 'f)`, reference has a longer lifetime than the data it references
```
I will try to make `map`s accept whatever the parser returns instead of `&mut I`https://github.com/winnow-rs/winnow/issues/131#issuecomment-2486637714 .

39555:
Updated with the new API.
- Without `_: &mut _`
- Without `as _`
```rust
use Infix::*;
expression(0, digit1.parse_to::<i32>())
    .prefix(dispatch! {any;
        '+' => Prefix(12, |_, a| Ok(a)),
        '-' => Prefix(12, |_, a: i32| Ok(-a)),
        _ => fail
    })
    .infix(dispatch! {any;
       '+' => Left(5, |_, a, b| Ok(a + b)),
       '-' => Left(5, |_, a, b| Ok(a - b)),
       '*' => Left(7, |_, a, b| Ok(a * b)),
       '/' => Left(7, |_, a, b| Ok(a / b)),
       '%' => Left(7, |_, a, b| Ok(a % b)),
       '^' => Right(9, |_, a, b| Ok(a ^ b)),
       _ => fail
    })
    .parse_next(i)

```

src/combinator/expression.rs
```diff
@@ -0,0 +1,340 @@
...
+use super::{empty, fail};
+
+/// Parses an expression based on operator precedence.
+#[doc(alias = "pratt")]
+#[doc(alias = "separated")]
+#[doc(alias = "shunting_yard")]
+#[doc(alias = "precedence_climbing")]
+#[inline(always)]
+pub fn expression<I, ParseOperand, O, E>(
+    start_power: i64,
```
epage, line 19: Should `start_power` also be a "builder" method?
- Only used in very specific cases
- Can better focus documentation on it

epage: Also, would referring to it as `precedence` or `precedence_level` be a more approachable user-facing term?

src/combinator/expression.rs
```diff
@@ -0,0 +1,340 @@
...
+            start_power: self.start_power,
+            parse_operand: self.parse_operand,
+            parse_prefix: parser,
+            parse_postfix: self.parse_postfix,
+            parse_infix: self.parse_infix,
+            i: Default::default(),
+            o: Default::default(),
+            e: Default::default(),
+        }
+    }
```
epage, line 87: For these, how come we don't do
```suggestion
    #[inline(always)]
    pub fn prefix<NewParsePrefix>(
        mut self,
        parser: NewParsePrefix,
    ) -> Expression<I, O, ParseOperand, NewParsePrefix, Post, Pix, E>
    where
        NewParsePrefix: Parser<I, Prefix<I, O, E>, E>,
    {
        self.parse_prefix = parser;
        self
    }
```

39555: Different types of the `Expression` due to the different type of the new parser

src/combinator/expression.rs
epage: When a user sets it, is it the same precedence level they set for their operator? If so, I'd use `current_` as the prefix

PR #618, opened 2024-11-14
PoC: Pratt parsing with `shunting yard` algorithm

39555:
Attempt №2 #614 

This is much smaller implementation based on the modified `shunting yard` from the https://github.com/bourguet/operator_precedence_parsing/tree/master

Differences from the previous Pratt implementation:
- No more recursion. The explicit `Vec` stack is now used, with one stack for operands and another for operators.
- Without `RefCell`s

Differences from the https://en.wikipedia.org/wiki/Shunting_yard_algorithm:
- Parsing is done in a single pass without first converting the expression to Polish notation.
- Braces '(' are handled as an `operand` using recursive sub-expression similar to https://github.com/rust-bakery/nom/pull/1362 (Wikipedia hardcodes braces into the algorithm itself)
  - The `operator_precedence_parsing` repository introduces special `prefix_action` and `postfix_action` mutable closures for handling braces. This complicates the algorithm, so it’s out of scope for our PoC. However, we’ll keep it in mind if recursion for braces is undesirable.

This is extremely barebones for now without the fancy UX we will agree later. it is 3 parsers slapped into the function signature: one each for prefix, postfix, and infix. Prefix and postfix parsers should return
`(power, &dyn Fn(O) -> O)` and the infix should return the ugly `(left_power, right_power, &dyn Fn(O) -> O)` just 2 powers for now without a trick with `Assoc` enum.

If this is the way to go I will apply your rewiew suggestions from #614 where they’re still applicable.

## Minor things to consider
- [ ] A user provided stack similar to `Accumulate`
- [ ] Error kinds. Algorithm has `missing operand` and `value left on stack`

EDIT: This algorithm is described as `The-Double-E-Method` in https://github.com/erikeidt/erikeidt.github.io/blob/master/The-Double-E-Method.md

### DIFF
--- a/src/combinator/shunting_yard.rs
+++ b/src/combinator/shunting_yard.rs
@@ -3,9 +3,11 @@ use crate::error::{ErrMode, ErrorKind, ParserError};
 use crate::stream::{Stream, StreamIsPartial};
 use crate::{PResult, Parser};
 
+use super::precedence::Assoc;
 use super::trace;
 
-pub fn precedence<'i, I, ParseOperand, ParseInfix, ParsePrefix, ParsePostfix, Operand: 'static, E>(
+pub fn precedence<I, ParseOperand, ParseInfix, ParsePrefix, ParsePostfix, Operand: 'static, E>(
+    start_precedence: i64,
     mut operand: ParseOperand,
     mut prefix: ParsePrefix,
     mut postfix: ParsePostfix,
@@ -14,13 +16,14 @@ pub fn precedence<'i, I, ParseOperand, ParseInfix, ParsePrefix, ParsePostfix, Op
 where
     I: Stream + StreamIsPartial,
     ParseOperand: Parser<I, Operand, E>,
-    ParseInfix: Parser<I, (usize, usize, &'i dyn Fn(Operand, Operand) -> Operand), E>,
-    ParsePrefix: Parser<I, (usize, &'i dyn Fn(Operand) -> Operand), E>,
-    ParsePostfix: Parser<I, (usize, &'i dyn Fn(Operand) -> Operand), E>,
+    ParseInfix: Parser<I, (Assoc, fn(&mut I, Operand, Operand) -> PResult<Operand, E>), E>,
+    ParsePrefix: Parser<I, (i64, fn(&mut I, Operand) -> PResult<Operand, E>), E>,
+    ParsePostfix: Parser<I, (i64, fn(&mut I, Operand) -> PResult<Operand, E>), E>,
     E: ParserError<I>,
 {
     trace("precedence", move |i: &mut I| {
         let result = shunting_yard(
+            start_precedence,
             i,
             operand.by_ref(),
             prefix.by_ref(),
@@ -31,7 +34,8 @@ where
     })
 }
 
-fn shunting_yard<'i, I, ParseOperand, ParseInfix, ParsePrefix, ParsePostfix, Operand: 'static, E>(
+fn shunting_yard<I, ParseOperand, ParseInfix, ParsePrefix, ParsePostfix, Operand: 'static, E>(
+    start_precedence: i64,
     i: &mut I,
     mut operand: ParseOperand,
     mut prefix: ParsePrefix,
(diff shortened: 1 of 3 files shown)


src/combinator/shunting_yard.rs
```diff
@@ -0,0 +1,218 @@
...
+    ParseOperand: Parser<I, Operand, E>,
+    ParseInfix: Parser<I, (usize, usize, &'i dyn Fn(Operand, Operand) -> Operand), E>,
+    ParsePrefix: Parser<I, (usize, &'i dyn Fn(Operand) -> Operand), E>,
+    ParsePostfix: Parser<I, (usize, &'i dyn Fn(Operand) -> Operand), E>,
+    E: ParserError<I>,
+{
+    // what we expecting to parse next
+    let mut waiting_operand = true;
+    // a stack for computing the result
+    let mut value_stack = Vec::<Operand>::new();
```
epage, line 52: We'll need to mark this as requiring `std`.  That is the one benefit to recursion that it can operate in `no_std` environments.

src/combinator/shunting_yard.rs
```diff
@@ -0,0 +1,218 @@
...
+            ),
+            trace(
+                "prefix",
+                dispatch! {any;
+                    '+' => trace("+", empty).value((9, (&|a| a) as _)),
+                    '-' => trace("-", empty).value((9, (&|a: i32| -a) as _)),
+                    _ => fail
+                },
+            ),
+            trace("postfix", fail),
```
epage, line 185: `precedence` could put these traces on the parameters it passes to `shunting_yard`

epage: Granted, encouraging users to do it makes the parameter list easier to read

src/combinator/shunting_yard.rs
```diff
@@ -0,0 +1,218 @@
...
+        ascii::digit1,
+        combinator::{cut_err, delimited, empty, fail, peek},
+        dispatch,
+        token::any,
+    };
+
+    use super::*;
+
+    fn parser(i: &mut &str) -> PResult<i32> {
+        // TODO: how to elide the closure type without ugly `as _`
```
epage, line 168: Its not clear to me what problem you are having here

src/combinator/shunting_yard.rs
epage: >  A user provided stack similar to Accumulate

Ah, curious idea to explore.  I wouldn't put this as a blocker but we can create an issue and see if it garners interest

src/combinator/shunting_yard.rs
```diff
@@ -0,0 +1,218 @@
...
+            break 'parse;
+        }
+    }
+
+    while let Some(op) = operator_stack.pop() {
+        evaluate(&mut value_stack, op);
+    }
+    // if eval_stack.len() > 1 {
+    //     // Error: value left on stack
+    // }
```
epage, line 106: >  Error kinds. Algorithm has missing operand and value left on stack

I can see it being important to know of a "missing operand".

What end-user condition leaves a value on the stack or is that more of an assert?

src/combinator/shunting_yard.rs
epage, line 1: > If this is the way to go I will apply your rewiew suggestions from https://github.com/winnow-rs/winnow/pull/614 where they’re still applicable.

I assume this API style of API, removing `RefCell` and allowing `dispatch!` could be applied to #614.

What is your overall impression of the two?

I'm also curious about the performance of recursion vs iteration but I suspect some differences, like use of `dispatch!` would bias things

src/combinator/shunting_yard.rs
```diff
@@ -0,0 +1,218 @@
...
+pub fn precedence<'i, I, ParseOperand, ParseInfix, ParsePrefix, ParsePostfix, Operand: 'static, E>(
+    mut operand: ParseOperand,
+    mut prefix: ParsePrefix,
+    mut postfix: ParsePostfix,
+    mut infix: ParseInfix,
+) -> impl Parser<I, Operand, E>
+where
+    I: Stream + StreamIsPartial,
+    ParseOperand: Parser<I, Operand, E>,
+    ParseInfix: Parser<I, (usize, usize, &'i dyn Fn(Operand, Operand) -> Operand), E>,
```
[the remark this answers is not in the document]
39555: Yes. The algorithm uses two powers for infix to determine what to parse next. From matklad's https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
```
expr:      A       +       B       +       C
power:  0      3      3.1      3      3.1     0
```
the enum in `chumsky` automatically bumps the value with some clever trick
```
impl Associativity {
    fn left_power(&self) -> u32 {
        match self {
            Self::Left(x) => *x as u32 * 2,
            Self::Right(x) => *x as u32 * 2 + 1,
        }
    }

    fn right_power(&self) -> u32 {
        match self {
            Self::Left(x) => *x as u32 * 2 + 1,
            Self::Right(x) => *x as u32 * 2,
        }
    }
}
```

src/combinator/shunting_yard.rs
39555: I will write a benchmark with dispatches and stripped down ReffCells and tuples.  We will see what is the best. The explicit stack is nice if the interface allows the user to customize the type such as VecDeque or SmallVec or something for no_std. Both functions are really similar except the recursion part

39555:
A really great description of this algorithm https://github.com/erikeidt/erikeidt.github.io/blob/master/The-Double-E-Method.md

Issue #609, opened 2024-10-20
`dispatch!` unable to handle cfg attributes


zeenix:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### rust version

1.81.0

### winnow version

0.6.13

### Minimal reproducible code

```rust
fn main() {
    use winnow::combinator::{dispatch, empty, fail, preceded};
    use winnow::prelude::*;
    use winnow::token::any;

    fn escaped(input: &mut &str) -> PResult<char> {
        preceded('\\', escape_seq_char).parse_next(input)
    }

    fn escape_seq_char(input: &mut &str) -> PResult<char> {
        dispatch! {any;
            'b' => empty.value('\u{8}'),
            #[cfg(unix)]
            '"' => empty.value('"'),
            _ => fail::<_, char, _>,
        }
        .parse_next(input)
    }

    assert_eq!(escaped.parse_peek("\\b\""), Ok(("\"", '\u{8}')));
}
```

### Steps to reproduce the bug with the above code

cargo run

### Actual Behaviour

```rust
   Compiling tmp-unpavh v0.1.0 (/home/zeenix/.cache/cargo-temp/tmp-UNpAvH)
error: no rules expected the token `#`
  --> src/main.rs:13:13
   |
13 |             #[cfg(unix)]
   |             ^ no rules expected this token in macro call
   |
note: while trying to match meta-variable `$pat:pat`
  --> /home/zeenix/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winnow-0.6.20/src/macros/dispatch.rs:43:30
   |
43 |     ($match_parser: expr; $( $pat:pat $(if $pred:expr)? => $expr: expr ),+ $(,)? ) => {
   |                              ^^^^^^^^

warning: unused imports: `empty` and `fail`
 --> src/main.rs:2:40
  |
2 |     use winnow::combinator::{dispatch, empty, fail, preceded};
  |                                        ^^^^^  ^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `winnow::token::any`
 --> src/main.rs:4:9
  |
4 |     use winnow::token::any;
  |         ^^^^^^^^^^^^^^^^^^

[...]

PR #619, opened 2024-11-14
fix(dispatch): handle `#[...]` attributes

39555:
Closes: #609 

Handle `#[...]` attributes in `dispatch!` macro


39555:
This is not related to the `Pratt` prs. I just resolved an issue that I knew how to fix by the way

epage:
I appreciate it!

I bet `seq!` could also benefit from this ;)

(no obligation and not a blocker for this PR; incremental improvement is great!)

PR #622, opened 2024-11-17
Pratt example

39555:
WIP. An example of the usage of the `Pratt` parser for parsing a weird cexpr.

 The result of the parsing is a nicely formatted ast and the expression in prefix notation.

 `1 + 2 + 3`:

```
ADD
  ADD
    VAL 1
    VAL 2
  VAL 3

(+ (+ 1 2) 3)
```

## Parser Problems

- [ ] Parsing any complex postfix operator `a ? b : c`, `foo()`, `foo(1 + 2)`, `a[1 + 2]`  cannot be done without parsing it inside the `operand` parser (which means breaking the prefix precedence). Maybe the api should provide the input inside the operator closures. But I'm having strange lifetime troubles with it for now.
- [ ] Maybe closures should return an error instead of plain value to allow validation of the input e.g dereferencing the literal `1->foo` or handling unbalanced delimiter in complex postfixes.
- [ ] Concept of `neither` associativity. I don't know yet how it works but the parser could potentially reject `a == b == c` somehow.


examples/pratt/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+                _ => fail
+            }
+        , multispace0)),
+        trace(
+            "postfix",
+            delimited(multispace0, alt((
+                dispatch! {any;
+                    '!' => empty.value((20, (|_: &mut _, a| Ok(Expr::Fac(Box::new(a)))) as _)),
+                    '?' => empty.value((3, (|i: &mut &str, cond| {
+                        let (left, right) = preceded(multispace0, cut_err(separated_pair(pratt_parser, delimited(multispace0, ':', multispace0), pratt_parser))).parse_next(i)?;
```
epage, line 108: Having to put  `multispace0`s in here means that we have to successfully parse more of the input before we find that it doesn't match, hurting performance.  I assume the way to handle this is to lex into tokens and then run this parser on tokens which will have the multi-space taken care of.

examples/pratt/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+                   '+' => empty.value((14, 15, (|_: &mut _, a, b| Ok(Expr::Add(Box::new(a), Box::new(b)))) as _)),
+                   '-' => alt((
+                        dispatch!{take(2usize);
+                            "ne" => empty.value((10, 11, (|_: &mut _, a, b| Ok(Expr::NotEq(Box::new(a), Box::new(b)))) as _)),
+                            "eq" => empty.value((10, 11, (|_: &mut _, a, b| Ok(Expr::Eq(Box::new(a), Box::new(b)))) as _)),
+                            "gt" => empty.value((12, 13, (|_: &mut _, a, b| Ok(Expr::Greater(Box::new(a), Box::new(b)))) as _)),
+                            "ge" => empty.value((12, 13, (|_: &mut _, a, b| Ok(Expr::GreaterEqual(Box::new(a), Box::new(b)))) as _)),
+                            "lt" => empty.value((12, 13, (|_: &mut _, a, b| Ok(Expr::Less(Box::new(a), Box::new(b)))) as _)),
+                            "le" => empty.value((12, 13, (|_: &mut _, a, b| Ok(Expr::LessEqual(Box::new(a), Box::new(b)))) as _)),
+                        _ => fail
```
epage, line 146: Is this missing indentation?

examples/pratt/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+    use super::*;
+    #[test]
+    fn test_simple() {
+        let r = pratt_parser
+            // .parse("+- --!&**foo! + 3 * 4 - bar ^ e")
+            // .parse("foo(a + b)")
+            // .parse("1 + 2 * *4^7! + 6")
+            .parse("foo(1 + 2 + 3) + bar() ? 1 : 2")
+            .unwrap();
+        println!("{r}");
```
epage, line 374: With `snapbox` we can do snapshot testing of the `.to_string()`

39555: I just have pushed a lot of tests. There is another api complication thing that currently fails. When invoking a recursive parser in a ternary operator, it should know its starting precedence. Consider:
`a ? b : c, d`
it should parse as `(, (? a b c) d)`. But currently it parses as `(? a b (, c d))`. The second part after `:` doesn't know it's part of the ternary operator. 2 solutions are:
- Allow users to provide a starting binding power e.g `precedence(0, ...)`. The user would call `precedence(ternary_precedence+1, ...)` inside the ternary operator.
- Require users to rebuild child parsers excluding operators with precedence lower than the current one.

examples/pratt/parser.rs
39555: Yes. Thanks. Fixed. It seems like rustfmt is having a hard time formatting macros.

examples/pratt/parser.rs
39555: When all the missing parts work, we will see the full interface and can consider the most user-friendly design

39555:
Well, the example is complete and all the features are there! I assume with #618 it would be the same. I may check it later.

epage:
@39555 I have to say, I am thoroughly impressed with the dedication you have put to this investigation, having
- Implemented the initial recursive version
- Implemented an iterative version out of my concern for stackoverflows
- Implemented a full C expression parser with it to see what feature are missing, adding features that I'm not seeing in other libraries that provide a generic "pratt" parser

ssmendon:
I don't know what the status is on this PR, but I was working on fixing this up for `v0.7` for my own crate.

You can check it out at https://github.com/ssmendon/rollers/tree/main/crates/pratt if it's helpful at all.

PR #625, opened 2024-11-22
docs(tutorial): Clarify examples in chapter 5

Nukesor:
While going through the tutorial, I found the description on the last example of chapter 5 very confusing.

As a beginner, at this point of the tutorial I had the mental model that a Parser always consumes/advdances the input.
Suddenly `take` is introduced without any further info, which uses checkpoints under the hood.

And to make this even more confusing, the output of the `parse_list` parser is changed to "accumulat[e] the results into a `()`", which really doesn't make a lot of sense if you don't know that `take` doesn't care about the output of the contained parser due to the checkpoint logic.

There's really a lot of concepts to unwrap in this small piece of code and it took a bit of research to figure out why stuff worked the way it does.

I hope the new text is able to convey this knowledge in a better way.

### DIFF
--- a/src/_tutorial/chapter_5.rs
+++ b/src/_tutorial/chapter_5.rs
@@ -136,8 +136,8 @@
 //! # }
 //! ```
 //!
-//! You'll notice that the above allows trailing `,` when we intended to not support that. We can
-//! easily fix this by using [`separated`]:
+//! You'll notice that the above allows trailing `,`. However, if that's not desired, it
+//! can easily be fixed by using [`separated`] instead of [`repeat`]:
 //! ```rust
 //! # use winnow::prelude::*;
 //! # use winnow::token::take_while;
@@ -199,9 +199,22 @@
 //! }
 //! ```
 //!
-//! If you look closely at [`repeat`], it isn't collecting directly into a [`Vec`] but
-//! anything that implements the [`Accumulate`] trait to gather the results. This lets us make more complex parsers than we did in
-//! [`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing the captured input:
+//! If you look closely at [`separated`] and [`repeat`], they aren't limited to collecting
+//! the result into a [`Vec`], but rather anything that implements the [`Accumulate`] trait.
+//! [`Accumulate`] is for instance also implemented for [`HashSet`], [`String`] and `()`.
+//!
+//! This lets us build more complex parsers than we did in
+//! [`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing
+//! the consumed input.
+//!
+//! `take` works by
+//! 1. Creating a [`checkpoint`][Stream::checkpoint]
+//! 2. Running the inner parser, in our case the `parse_list` parser, which will advance the input
+//! 3. Returning the slice from the first checkpoint to the current position.
+//!
+//! Since the result of `parse_list` gets thrown away, we
+//! accumulates into a `()` to not waste work creating an unused `Vec`.
+//!
 //! ```rust
 //! # use winnow::prelude::*;
 //! # use winnow::token::take_while;
@@ -274,7 +287,9 @@ use crate::combinator;
 use crate::combinator::repeat;
 use crate::combinator::separated;
 use crate::stream::Accumulate;
+use crate::stream::Stream;
 use crate::Parser;
+use std::collections::HashSet;
 use std::vec::Vec;
 
 pub use super::chapter_4 as previous;


src/_tutorial/chapter_5.rs
```diff
@@ -199,9 +199,23 @@
 //! }
 //! ```
 //!
-//! If you look closely at [`repeat`], it isn't collecting directly into a [`Vec`] but
-//! anything that implements the [`Accumulate`] trait to gather the results. This lets us make more complex parsers than we did in
-//! [`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing the captured input:
+//! If you look closely at [`separated`] and [`repeat`], they don't always collect the result into
+//! a [`Vec`], but rather anything that implements the [`Accumulate`] trait.
```
epage, line 203:
```suggestion
//! If you look closely at [`separated`] and [`repeat`], they aren't limited to collecting the result into
//! a [`Vec`], but rather anything that implements the [`Accumulate`] trait.
```

src/_tutorial/chapter_5.rs
```diff
@@ -199,9 +199,23 @@
...
-//! If you look closely at [`repeat`], it isn't collecting directly into a [`Vec`] but
-//! anything that implements the [`Accumulate`] trait to gather the results. This lets us make more complex parsers than we did in
-//! [`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing the captured input:
+//! If you look closely at [`separated`] and [`repeat`], they aren't limited to collecting
+//! the result into a [`Vec`], but rather anything that implements the [`Accumulate`] trait.
+//! [`Accumulate`] is for instance also implemented for [`HashSet`], [`String`] and `()`.
+//!
+//! This enables you to build more complex parsers than shown in [`chapter_2`] by accumulating
+//! complex expressions into a single `&str`. This is done by [`take`][Parser::take]-ing the
+//! `parse_list` parser, which effectively encapsulates it in a [`Take`].
```
epage, line 208: Saying we accumulate into a `&str` is inaccurate.  I'd prefer to keep the old statement.

Also, talking about `Take` isn't too meaningful as it only exists because we couldn't `-> impl Parser` in a trait.  We now can once we bump our MSRV to 1.75 and are willing to make a breaking change.

epage:
Thank you for taking the time to fix this to help others in the future!

src/_tutorial/chapter_5.rs
Nukesor: You're right, just checked and it reads it into a slice of the respective stream type.
https://docs.rs/winnow/latest/src/winnow/combinator/parser.rs.html#666

The old statement is also a quite confusing though:
```
//! If you look closely at [`repeat`], it isn't collecting directly into a [`Vec`] but
//! anything that implements the [`Accumulate`] trait to gather the results. This lets us make more complex parsers than we did in
//! [`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing the captured input:
 ```
 
 It talks about accumulating into `()` and taking the captured output.
 But that's very much confusing as it sounds like you would accumulate the data into a `()` that's then captured, which that doesn't make a lot of sense. There's no captured output per-se.
 
 ---
 
Anyhow, should I close the PR then and wait until the bump to `MSRV` to 1.75 has been done :)?

Comment (epage) [reply src/_tutorial/chapter_5.rs:208]: > It talks about accumulating into () and taking the captured output.
But that's very much confusing as it sounds like you would accumulate the data into a () that's then captured, which that doesn't make a lot of sense. There's no captured output per-se.

The old phrase is
```md
This lets us make more complex parsers than we did in
[`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing the captured input:
```
I feel like the key word that is off here is "captured".  Looking at the docs, we tend to focus on "consumed" and I think that is an improvement here:
```md
This lets us make more complex parsers than we did in
[`chapter_2`] by accumulating the results into a `()` and [`take`][Parser::take]-ing the consumed input:
```


> Anyhow, should I close the PR then and wait until the bump to MSRV to 1.75 has been done :)?

No, we just frame things around `Parser::take` and ignore `Take`.

Nukesor: Ah that makes sense, thanks for the clarification :)

Also good suggestion to change `captured` -> `consumed`. That makes this sentence a lot more clear to me.

src/_tutorial/chapter_5.rs
```diff
@@ -199,9 +199,23 @@
...
+//! the consumed input.
+//!
+//! `take` works by creating a [`checkpoint`][Stream::checkpoint] before running the
+//! inner parser. In our case the `parse_list` parser then advances to the end of the digit list,
+//! afterwards `take` reads the offset (i.e. how many chars were matched by `parse_list`),
+//! restores the checkpoint, and returns the whole digit list as a single string based on
+//! the previously read offset.
+//!
+//! Due to this `take` doesn't care about the result of `parse_list`, which is why `parse_list`
+//! simply accumulates into a `()`, which effectively discards the product of `parse_list`.
```
epage, line 217: The sentence starting with "In our case" feels a bit off.  Also, I don't think we need to talk about offsets (and that invites going even deeper into the implementation weeds).  In trying to come up with an alternative, it made me wonder if a list would work better.

```suggestion
//! `take` works by
//! 1. Creating a [`checkpoint`][Stream::checkpoint]
//! 2. Running the inner parser, in our case the `parse_list` parser, which will advance the input
//! 3. Creating a [`checkpoint`][Stream::checkpoint] 
//! 4. Returning the slice from the first checkpoint to the second in place of the inner parser's result
//! 
//! Since the result of `parse_list` gets thrown away, we 
//! accumulates into a `()` to not waste work creating an unused `Vec`.
```

Thoughts?

Nukesor: Looks good to me, makes it a lot easier to read and keep track of the steps.

Nukesor: I assume you decided to create a second checkpoint for the sake of simplicity?

The code doesn't work that way and resets to the checkpoint before advancing again by the offset of `parse_list` and returning the slice.

epage: I wanted to gloss over the reset.

We could probably drop the second checkpoint and say we return the slice from the checkpoint to the current location.

Nukesor: Yeah, I think I like that better :)
Adjusted it accordingly

PR #626, opened 2024-11-22
feat: Add `Parser::value_with` convenience method

emilyyyylime:
I hope this is simple enough to not warrant an issue, this PR adds a method `value_with` to `Parser` that takes a generator function `FnMut() -> O2` and returns a new parser much like `p.value(x)` but without the bound `O2: Clone`.

This also lets us replace the implementation of `.default_value()` to be `ValueWith::new(self, Default::default)`, but it would also change the signature to return `ValueWith<Self, impl Fn() -> O2, I, O, O2, E>`, so I figured I won't change that yet.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -178,6 +178,31 @@ pub trait Parser<I, O, E> {
         Value::new(self, val)
     }
 
+    /// Produce a value by calling the provided generator function
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use winnow::{error::ErrMode,error::ErrorKind, error::InputError, Parser};
+    /// use winnow::ascii::alpha1;
+    /// # fn main() {
+    ///
+    /// let mut parser = alpha1.value_with(|| 1234);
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -178,6 +178,31 @@ pub trait Parser<I, O, E> {
...
+    /// Produce a value by calling the provided generator function
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use winnow::{error::ErrMode,error::ErrorKind, error::InputError, Parser};
+    /// use winnow::ascii::alpha1;
+    /// # fn main() {
+    ///
+    /// let mut parser = alpha1.value_with(|| 1234);
```
epage, line 190: What separates this from the following?
```suggestion
    /// let mut parser = alpha1.map(|_| 1234);
```

emilyyyylime: I've got nothing to say except—yeah, that's a very good point. Closing :^)

PR #627, opened 2024-11-26
impl ParseSlice trait for Located

redforks:
To use `Parser::parse_to()` method on Located stream, I add the `ParserSlice` trait on Located.

It is a minor change, I skipped make an issue

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -2684,6 +2684,16 @@ impl<'a, R: FromStr> ParseSlice<R> for &'a str {
     }
 }
 
+impl<I, R> ParseSlice<R> for Located<I>
+where
+    I: ParseSlice<R>,
+{
+    #[inline(always)]
+    fn parse_slice(&self) -> Option<R> {
+        self.input.parse_slice()
+    }
+}
+
 /// Convert a `Stream` into an appropriate `Output` type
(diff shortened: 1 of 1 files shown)


src/stream/mod.rs
```diff
@@ -2684,6 +2684,16 @@ impl<'a, R: FromStr> ParseSlice<R> for &'a str {
...
 
+impl<I, R> ParseSlice<R> for Located<I>
+where
+    I: ParseSlice<R>,
+{
+    #[inline(always)]
+    fn parse_slice(&self) -> Option<R> {
+        self.input.parse_slice()
+    }
+}
```
epage, line 2695: > To use Parser::parse_to() method on Located stream, I add the ParserSlice trait on Located.

You don't need to implement `ParseSlice` to use `parse_to` with a `Located<I>`
```rust
    fn parse_to<O2>(self) -> ParseTo<Self, I, O, O2, E>
    where
        Self: core::marker::Sized,
        I: Stream,
        O: ParseSlice<O2>,
        E: ParserError<I>,
```
`ParseSlice` must be implemented for the `O` or output of a parser, not for the `I` or input, like `Located`.

The design of winnow is centered on separating the stream from slices and isolating slice-like operations to slices.

PR #628, opened 2024-12-06
A couple of typo fixes

TheGiraffe3:
My apologies for not opening an issue about these, but I didn't think they warranted discussion, as they're obviously wrong.

I found these with a different spellcheck.


epage (approved):
Thanks!

Issue #634, opened 2024-12-16
impl Compare<&str> and Compare<&[u8]> for AsRef<[u8]>


soulstompp:
I noticed that a handful of impls of Compare<T> could be eliminated by handling AsRef<[u8]> more generically. I feel like the code is ever-so slightly more manageable this way. I wanted to open this issue to allow for discussion and will relate my pull request to this issue here shortly. 

What I wasn't expecting was it seems to have helped out with performance across the majority of benchmarks and left everything else neutral. I assume the improvement is because &str (which I know causes some performance issues in some instances) is now being compared as &[u8] even when both sides of the &str comparisons. 

These are the benchmark results I got after running against main and then my branch:

```
t not found, using plotters backend
http/complete/291       time:   [628.45 ns 632.36 ns 637.42 ns]
                        thrpt:  [435.38 MiB/s 438.86 MiB/s 441.59 MiB/s]
                 change:
                        time:   [-3.9294% -3.5841% -3.2019%] (p = 0.00 < 0.05)
                        thrpt:  [+3.3079% +3.7173% +4.0901%]
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
http/streaming/291      time:   [739.42 ns 739.92 ns 740.46 ns]
                        thrpt:  [374.79 MiB/s 375.07 MiB/s 375.32 MiB/s]
                 change:
                        time:   [-5.7336% -5.3313% -4.9403%] (p = 0.00 < 0.05)
                        thrpt:  [+5.1970% +5.6315% +6.0823%]
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe

     Running examples/ini/bench.rs (target/release/deps/ini-43021e6b6bb897e2)
Gnuplot not found, using plotters backend
ini/bytes/110           time:   [421.09 ns 422.31 ns 423.56 ns]
                        thrpt:  [247.67 MiB/s 248.41 MiB/s 249.13 MiB/s]
                 change:
                        time:   [-11.524% -10.950% -10.450%] (p = 0.00 < 0.05)
                        thrpt:  [+11.669% +12.296% +13.025%]
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
ini/str/110             time:   [458.08 ns 459.91 ns 461.95 ns]
[...]

PR #635, opened 2024-12-16
impl Compare<&str> and Compare<&[u8]> for AsRef<[u8]>

soulstompp:
Impl AsRef<[u8]> for &str and &[u8]  as mentioned in issue #634. This makes the code slightly more manageable and helps with performance.


soulstompp:
I will get the commit message fixed to make the linter happy but not sure what the proper description should be since it seemed like a `chore` but might be a  `perf(...)`.

epage:
As this needs more design discussion in #634, I'm marking this as a Draft for now

PR #640, opened 2024-12-30
chore(deps): bump anstream to latest version

orhun:
Fixes [`RUSTSEC-2024-0404`](https://rustsec.org/advisories/RUSTSEC-2024-0404.html)

[The changelog](https://github.com/rust-cli/anstyle/blob/main/crates/anstream/CHANGELOG.md) does not seem to contain anything breaking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,29 +20,14 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.3.2"
+version = "0.6.18"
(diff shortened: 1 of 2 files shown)


epage:
In libraries, I try to keep the version requirements to a minimum.  As this unsoundness issue does not affect the development of `winnow`, I lean towards leaving it to dependents to decide whether they are affected and should upgrade.

PR #666, opened 2025-01-03
fix(stream)!: Change ContainsToken from Fn to FnMut

### LINKED ISSUE #558: accept `FnMut` instead of just `Fn` as `take_till` parameter

### winnow version

0.6.13

### Describe your use case

I was trying to write a parser that runs `take_till` until it finds a matching bracket (keeping all the brackets found inside balanced), which requires a mutable counter variable to keep track of how many brackets have been found so far

Allow passing `FnMut` to `take_till` to allow for more complex parsers without needing to make the entire parser stateful or juggling data in and out of a `Cell`

I've used `Cell` and a `move` closure to keep the closure `Fn` while still allowing for mutating the counter, but I feel like it makes the closure unnecessarily complicated and long just to deal with juggling data in and out of the `Cell`, so I think if there's no functional reason to constraint the parameter to `Fn`, it should be relaxed

### Additional Context

https://github.com/winnow-rs/winnow/discussions/552#discussioncomment-10032721

epage:
Fixes #558

### DIFF
--- a/examples/arithmetic/parser_lexer.rs
+++ b/examples/arithmetic/parser_lexer.rs
@@ -71,28 +71,28 @@ pub enum Oper {
 
 impl winnow::stream::ContainsToken<Token> for Token {
     #[inline(always)]
-    fn contains_token(&self, token: Token) -> bool {
+    fn contains_token(&mut self, token: Token) -> bool {
         *self == token
     }
 }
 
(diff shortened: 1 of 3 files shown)


epage:
We may want to revert this and re-weigh whether it is worth it.  It produces a lot of warnings when calling the trait on constants and they can't be changed to `static` due to ranges being non-copyable.

Issue #297, opened 2023-07-28
Failable `Accumulate`


epage:
### Please complete the following tasks

- [X] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [X] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue+label%3AS-wont-fix+is%3Aclosed) issues

### winnow version

0.5

### Describe your use case

If you want to `Accumulate` into a `HashMap`, the only policy we can currently implement is last instance wins.  What if instead someone wants duplicate keys to error?

Similarly, what about supporting allocation errors?

### Describe the solution you'd like

Update the trait for handle failable accumulate operations

epage:
In thinking on this, we are unlikely to implement `Accumulate` for a failable type and most people are likely to get a failable type from a dependency, meaning they can't implement `Accumulate` for it themself.  So what would be more useful is a `try_fold`.

As such, I'm going to close this out.  We can always reconsider if a strong enough need arises.

PR #685, opened 2025-01-10
feat(combinator): try_fold and verify_fold

### LINKED ISSUE #99: Failable `fold` support (ie `try_fold`)

### winnow version

0.2.0

### Describe your use case

Sometimes the container you are writing to as part of `fold` can fail
- If you are checking for allocation errors with `Vec` (see rust-bakery/nom#1629)
- Like with `toml_edit` where we have to enforce rules about when you can extend tables

A `try_fold` that allows the operation to fail

_No response_

### Additional Context

_No response_

Altair-Bueno:
Closes #99

### Implementation notes

Not sure if the documentation examples are good enough. In particular, I couldn't find any (non-nightly) simple examples for showcasing `try_fold`. Should we add something a bit more verbose or keep things simple?

### DIFF
--- a/src/combinator/multi.rs
+++ b/src/combinator/multi.rs
@@ -3,6 +3,7 @@
 use crate::combinator::trace;
 use crate::error::ErrMode;
 use crate::error::ErrorKind;
+use crate::error::FromExternalError;
 use crate::error::ParserError;
 use crate::stream::Accumulate;
 use crate::stream::Range;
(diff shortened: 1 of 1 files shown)


src/combinator/multi.rs
```diff
@@ -289,6 +290,165 @@ where
...
+    /// }
+    ///
+    /// assert_eq!(parser("abc"), Ok(("", b"abc".to_vec())));
+    /// assert_eq!(parser("abc123"), Ok(("123", b"abc".to_vec())));
+    /// assert_eq!(parser("123123"), Ok(("123123", vec![])));
+    /// assert_eq!(parser(""), Ok(("", vec![])));
+    #[doc(alias = "try_fold_many0")]
+    #[doc(alias = "try_fold_many1")]
+    #[doc(alias = "try_fold_many_m_n")]
+    #[doc(alias = "try_fold_repeat")]
```
epage, line 425: ditto

src/combinator/multi.rs
```diff
@@ -289,6 +290,165 @@ where
...
+    #[inline(always)]
+    pub fn try_fold<Init, Op, OpError, Result>(
+        mut self,
+        mut init: Init,
+        mut op: Op,
+    ) -> impl Parser<Input, Result, Error>
+    where
+        Init: FnMut() -> Result,
```
epage, line 433: Should we allow `Init` to fail?

src/combinator/multi.rs
```diff
@@ -1353,3 +1513,135 @@ where
 
     Ok(acc)
 }
+
+#[inline(always)]
+fn verify_fold_impl<I, O, E, F, G, H, R>(
```
epage, line 1518: Can you name this like `repeat().fold()`, so it'd be `verify_fold_m_n`

src/combinator/multi.rs
```diff
@@ -1353,3 +1513,135 @@ where
...
+            }
+            Err(e) => return Err(e),
+        }
+    }
+
+    Ok(acc)
+}
+
+#[inline(always)]
+fn try_fold_impl<I, O, E, F, G, H, R, GE>(
```
epage, line 1583: Can you name this like `repeat().fold()`, so it'd be `try_fold_m_n`

src/combinator/multi.rs
Altair-Bueno: > Should we allow Init to fail?

I think it's unrealistic a collection/accumulator would fail on creation. Either way a user could do something like this:

```rs
let foo = Foo::new()?;

combinator::repeat(.., some_parser).try_fold(||foo.clone(), |f,n|todo!("Do something with f"))
```

Note that `std::iter::Iterator::try_fold` does not work that way, so i think we are on our own: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold

epage: Failable allocators is the use case I'm thinking of, see #297.

epage: To clarify, `Vec::new()` wouldn't fail but `Vec::with_capacity(n)` would

Altair-Bueno: The only downside I can think of is that users would have to write

```rs
combinator::repeat(.., some_parser).try_fold(||Ok(Foo::default()), |f,n|todo!("Do something with f"))
```

Rather than


```rs
combinator::repeat(.., some_parser).try_fold(Foo::default, |f,n|todo!("Do something with f"))
```

Bit pedantic but definetly not the end of the world

epage: Eh, let's hold off for now and see what input we get.

epage:
btw I'm generally fine having commits squashed during review and much prefer the commits to match user-facing changes (so a `fix` meaning something in `master` is fixed, not fixed within the same PR)

PR #696, opened 2025-01-14
docs: Document ErrorKind variants

Nukesor:
While working on some manual error handling, I stumbled over the `ErrorKind` enum several times.

Since there's no real documentation in there, I thought it might be nice to have each of the variants explained, as most of them are only used in very specific circumstances that aren't immediately apparent to users.


Nukesor:
Closing in favor of removing the `ErrorKind` entirely. See #697

PR #717, opened 2025-01-27
docs(topics): Discuss nom's use of GATs

epage:

### DIFF
--- a/src/_topic/nom.rs
+++ b/src/_topic/nom.rs
@@ -71,6 +71,15 @@
 //! their behavior accordingly.
 //! See [partial] special topic.
 //!
+//! ### Parsers return [`Stream::Slice`], rather than [`Stream`]
+//!
+//! In `nom`, parsers like [`take_while`] parse a [`Stream`] and return a [`Stream`].
+//! When wrapping the input, like with [`Stateful`],
+//! you have to unwrap the input to integrate it in your application,
+//! requires [`Stream`] to be `Clone` (which requires `RefCell` for mutable external state),
+//! and is then expensive to `clone()`.
+//! Instead, [`Stream::Slice`] was added to track the intended type for parsers to return.
+//!
 //! ### `&mut I`
 //!
 //! For an explanation of this change, see [Why `winnow`][super::why]
@@ -122,4 +131,6 @@ use crate::_topic::partial;
 use crate::error::ErrMode;
 use crate::error::ModalResult;
 use crate::stream::Partial;
+use crate::stream::Stateful;
 use crate::stream::Stream;
+use crate::token::take_while;
--- a/src/_topic/nom.rs
+++ b/src/_topic/nom.rs
@@ -54,7 +54,7 @@
 //! - [gitoxide](https://github.com/Byron/gitoxide/pull/956) (gradual migration from nom
 //!   to winnow 0.5)
 //!
-//! ## API differences
+//! ## Differences
 //!
 //! ### Renamed APIs
 //!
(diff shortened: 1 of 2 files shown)


src/_topic/nom.rs
```diff
@@ -64,16 +64,75 @@
...
-//! ### Partial/streaming parsers
+//! ### GATs
 //!
-//! `nom` differentiated some parsers by being `streaming` or `complete`.
+//! `nom` v8 back-propagates how you will use a parser to parser functions using a language feature
+//! called GATs.
+//! Winnow avoids this.
+//!
+//! Benefits for avoiding GATs:
+//! - Predictable performance as writing; idiomatic `fn(&mut I) -> Result<O>` parser sever the
```
Hywan, line 74: What do you mean by “predictable performance”? Do you have a source to this statement?

src/_topic/nom.rs
```diff
@@ -64,16 +64,75 @@
...
+//!   back-propagation from GATs.
+//! - No "eek out X% perf improvement" pressure to contort a parser to be written declaratively
+//!   that is better written imperatively
```
Hywan, line 77: I don't understand what you're saying here. How declarative vs imperative has an impact on performance? In which scale? Also, why imperative is better than declarative?

src/_topic/nom.rs
```diff
@@ -64,16 +64,75 @@
...
+//! - Built-in parsers serve are simple examples of idiomatic parsers
```
Hywan, line 78: I don't understand how it relates to GAT, can you expand please?

src/_topic/nom.rs
```diff
@@ -64,16 +64,75 @@
...
+//! - Faster build times and smaller binary size as parsers only need to be generated for one mode, not upto 6
```
Hywan, line 79: It's actually 8 modes (check/emit x output/error x streaming/complete).
Also, this new mechanism helps to eliminate many cases/scenarios, which may simplify the generated code. It is true that it might slow build, but the binary size might not be larger. Do you have any benchmark supporting this claim?

src/_topic/nom.rs
```diff
@@ -64,16 +64,75 @@
...
 //! their behavior accordingly.
 //! See [partial] special topic.
 //!
+//! #### Eliding Output
+//!
+//! `nom` v8 back-propagates whether an Output will be used and skips its creation.
+//! For example, `value(Null, many0(_))` will avoid creating and pushing to a `Vec`.
+//! Previously, users had to select `count_many0` over `many0` to avoid creating a `Vec`.
+//! Instead, `repeat` returns an `impl Accumulate<T>` which could be a `Vec`, a `usize` for `count`
+//! variants, or `()` to do no extra work.
```
Hywan, line 99: I read that as an argument in favor of GAT, is it correct?

src/_topic/nom.rs
epage: I tried to clarify what I mean but it seems I need to further expand on this.

If you are writing a parser, you can make an innocuous looking change like replacing a built-in parser with your own to get some special logic you need. This can cause a surprising slowdown that is hard to identify by inspection because you severed the propagation of parser modes and every parser you call will be compiled in the least aggressive parser mode.

src/_topic/nom.rs
epage: This isn't trying to give trade offs but explain how each library handles a problem. Earlier, I did call out that GATs can offer better performance.

Both can avoid allocations. Purely declarative nom (or with trait impls) can avoid allocations in more cases as well as other work (depending on how well the optimizer works on Winnow)

src/_topic/nom.rs
epage: I've used nom's built-in parsers as examples of how to do things and have encouraged that for Winnow users.

Now with GATs, nom's built-io parsers are not the simple straightforward `fn(I) -> Result<I, O>` parser functions but wrappers around complex trait functions.

src/_topic/nom.rs
epage: See https://www.reddit.com/r/rust/comments/1ib8f1f/parser_combinators_with_nom_8_are_here/m9ir73e/ where switching from `fn(I) -> Result<I, O>` to trait impls made the parser go from 80KiB to 106KiB.

These numbers match the theory around monomorphization. Before we had one function that would work in 4 different ways. Now we have 4 functions, one the same as the previous and 3 with different parts trimmed out. The default version of the function isn't any different to allow stuff to be trimmed out. This is setting aside complete vs streaming but that should be neutral as that was two explicit functions and the one explicit function now should be monomorphized back to those two.

epage:
Note I replied directly to see if there is more clarification needed before updating the comparison.

epage:
I've made an attempt to improve the wording and organization in #731.

PR #720, opened 2025-01-28
feat(stream)!: Add `TokenSlice` to make parsing lexed tokens easier

### LINKED ISSUE #590: Use references for `&[T]`s token type
Let users decide if they want to clone or not when they use it

where an &T is returned instead of a T:
[code](https://docs.rs/winnow/latest/src/winnow/stream/mod.rs.html#717)

### LINKED ISSUE #637: Mark parsing higher order tokens easier. (perhaps implement Compare for &[T] where T:Compare ?)

### winnow version

0.6.20

### Describe your use case

I'm doing a small codecraft project with a language interpreter, and the design has multiple layers of parsing - string to tokens, and tokens to ast, etc.

It all works fine, but it is not very ergonomic:

To match [Token:X, Token:Y] I seem to need to do:

```
(
    one_of(|t| matches!(t, Token::X(_))).map(|t| match t {
                Token::X(s) => s.to_string(),
                _ => unreachable!(),
            }),
    one_of(|t| matches!(t, Token::Y)),
)
```

The ergonomics vs parsing strings are very vivid - there I could do 

(any, "Y") or use literal() etc.

I cannot implement Compare, which literal depends on, as it hits the foreign type rule. For one_of, ContainsToken is sufficient.

```

pub fn test(tokens: Vec<Token>) -> anyhow::Result<()> {
    use winnow::token::literal;
    let mut p = literal(&[Token::Var][..]);
    p.parse(&tokens[..])
        .map(|_| ())
        .map_err(|e: ParseError<_, ContextError>| anyhow!("failed {e:?}"))
}

pub fn test2(tokens: Vec<Token>) -> anyhow::Result<()> {
    use winnow::token::one_of;
    let mut p = one_of(&[Token::Var][..]);
    p.parse(&tokens[..])
        .map(|_| ())
        .map_err(|e: ParseError<_, ContextError>| anyhow!("failed {e:?}"))
}
```

What I would like to be able to write things like
```
(Token::If, Token::LeftParen, expr, Token::RightParen, statement).parse_next(i)
```

See the use case

_No response_

### Additional Context

_No response_

epage:
I was hoping to get `ContainsToken` support but I'm not seeing a reasonable
way of restructuring the traits without being blocked by the compiler
(overlapping impls) or having other side effects (less ergonomic for
other users).

`Compare` gets us the basics though.  The `Fn(&Token<'_>) -> bool`
variant of `ContainsToken` should still work and people can add their
own `Compare`s to get tuples working.

Also, didn't see a way to get a `impl Parser for TokenKind` to work
for similar reasons.

Location tracking is inspired by #512 and #591

Fixes #637
Fixes #590

### DIFF
--- a/src/token/tests.rs
+++ b/src/token/tests.rs
@@ -2460,9 +2460,7 @@ Ok(
         str![[r#"
 Err(
     Incomplete(
-        Size(
-            2,
-        ),
+        Unknown,
     ),
 )
 
@@ -2570,9 +2568,7 @@ Ok(
         str![[r#"
 Err(
     Incomplete(
-        Size(
-            2,
-        ),
+        Unknown,
     ),
 )
 
@@ -2820,13 +2816,6 @@ fn tokenslice_literals() {
         LeftCurly,
         RightCurly,
         Value,
-    }
-
-    impl SliceLen for TokenKind {
-        #[inline(always)]
-        fn slice_len(&self) -> usize {
-            1
-        }
     }
 
     impl<'i, 't> Parser<TokenSlice<'i, 't>, &'i Token<'t>, ErrMode<InputError<TokenSlice<'i, 't>>>>
--- a/src/token/tests.rs
+++ b/src/token/tests.rs
@@ -2800,17 +2800,15 @@ fn tokenslice_literals() {
         raw: &'i str,
     }
 
-    impl<'t> ContainsToken<&Token<'t>> for &str {
-        #[inline(always)]
-        fn contains_token(&self, token: &Token<'t>) -> bool {
-            *self == token.raw
+    impl PartialEq<&str> for Token<'_> {
+        fn eq(&self, other: &&str) -> bool {
+            self.raw == *other
         }
     }
 
-    impl<'t> ContainsToken<&Token<'t>> for TokenKind {
-        #[inline(always)]
-        fn contains_token(&self, token: &Token<'t>) -> bool {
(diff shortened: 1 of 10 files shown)


src/token/tests.rs
```diff
@@ -2787,3 +2785,111 @@ Ok(
...
+        Token {
+            kind: LeftParen,
+            raw: "(",
+        },
+        [
+            Token {
+                kind: Value,
+                raw: "hello",
+            },
+        ],
```
tisonkun, line 2884: I may expect here we have bare token instead of a list. Could you explain why we get a list here?

FYI, I'm trying to use `TokenSlice` to integrate with logos at https://github.com/cratesland/spath/pull/20. Hopefully I can finish it and serve it as a real-world example :P

tisonkun: Seems a default impl at:

```rust
impl<'s, I, E: ParserError<I>> Parser<I, <I as Stream>::Slice, E> for &'s str
where
    I: Compare<&'s str> + StreamIsPartial,
    I: Stream,
{
    #[inline(always)]
    fn parse_next(&mut self, i: &mut I) -> Result<<I as Stream>::Slice, E> {
        crate::token::literal(*self).parse_next(i)
    }
}
```

while TokenSlice's Slice is `&[T]`.

epage: Yes, its a bit unfortunate but couldn't think of a good alternative.

PR #725, opened 2025-01-29
fix(error): Add back ModalError functions for easier transition

epage:

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -137,6 +137,22 @@ impl<E> ErrMode<E> {
         matches!(self, ErrMode::Incomplete(_))
     }
 
+    /// Prevent backtracking, bubbling the error up to the top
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -149,14 +165,24 @@ impl<E> ErrMode<E> {
...
     pub fn convert<F>(self) -> ErrMode<F>
     where
         E: ErrorConvert<F>,
     {
         ErrorConvert::convert(self)
     }
+
+    /// Unwrap the mode, returning the underlying error
+    ///
+    /// Returns `None` for [`ErrMode::Incomplete`]
```
tisonkun, line 178: Returns Err?

epage: Thanks for pointing that out!  Fixed in 70eed322

PR #730, opened 2025-01-30
perf(token): Specialize behavior for `one_of`

epage:


epage:
Was hoping cutting down the layers (and doing checks earlier) could speed this up but the numbers look to be about the same.

There still might be a benefit to a `next_token_if` to make hand written parser functions easier but that is likely for another day.

PR #743, opened 2025-02-14
tests: update snapbox fixtures for should-panic-in-debug tests

decathorpe:
It looks like since more tests were migrated to snapbox, snapshots for some tests were just never generated - looks like it's mostly for tests that are expected to panic with `debug_assertions`, so they panic before hitting the assertion, but fail when run in without debug assertions because of the missing inline snapshot.

For this PR I simply ran `SNAPSHOTS=overwrite cargo test --release` to generate the missing snapshots. I have verified that the tests now pass in both debug and release mode.

I noticed this because we run package builds in Fedora with a profile derived from "release" (i.e. without debug assertions).


decathorpe:
(noticed that rustfmt was unhappy, ran `cargo fmt` and pushed again)

decathorpe:
Thanks! 🥳

PR #777, opened 2025-04-30
feat: Dynamically add context to Context

### LINKED ISSUE #744: Support adding context to `ContextError` based on arrays during runtime.

- [x] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [x] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue%20is%3Aclosed%20reason%3Anot-planned) issues

### winnow version

0.7.2

### Describe your use case

I'm currently writing a parser that detects a rather large set of keywords.
The keywords are represented as multiple enum types, that all implement `FromStr`.
Multiple enums were chosen it makes it easier to look at certain sets of keywords in the scope of a specific topic. 

When the parser cannot match any of the keywords, an error should be thrown that shows the list of all available keywords in the current scope. To do this programatically, the idea is to use strum's [`VariantNames`](https://docs.rs/strum_macros/latest/strum_macros/derive.VariantNames.html). This provides an easy and most importantly reliable automatic way of concatenating all keywords of the utilized keywords. Manual solutions are likely to go out of sync over time, as one has to adjust multiple error messages and go through a rather large list of keywords to keep them in sync.

In practice the error message creation would look something like this:
```rust
let mut concatenated = [
    SourceKeyword::VARIANTS,
    MetaKeyword::VARIANTS,
    RelationKeyword::VARIANTS,
]
.concat()
.join(", ");
concatenated.insert_str(0, "Expected one of the allowed properties: [");
concatenated.push(']');
```

The discussion started over here in a different issue:
https://github.com/winnow-rs/winnow/issues/496#issuecomment-2654165445

Right now it's possible to create a custom StringContext type in the likes of this:

```rust
#[derive(Clone, Debug)]
pub struct StringContext(String);

impl StringContext {
    pub fn new(value: impl Into<String>) -> Self {
        Self(value.into())
    }
}

impl std::fmt::Display for StringContext {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.0)
    }
}
```

Which can be called and used like this:

```rust
    fn keyword_parser(input: &mut &str) -> ModalResult<MyKeyword,  ContextError<StringContext>> {
```

This works fine as long as one works with `ContextError`s. But as soon as `parse` instead of `parse_next` is called, which returns a `ParseError` instead, this is no longer compatible:

`Display` is implemented for `ParseError`, with `E` being the inner error, which is the `ContextError<StringContext>` in this case.

```rust
impl<I, E> core::fmt::Display for ParseError<I, E>
where
    I: AsBStr,
    E: core::fmt::Display,
{
```

Currently, `Display` is only implemented for `ContextError<StrContext>`:
```rust
impl crate::lib::std::fmt::Display for ContextError<StrContext> {
```

Since it's a foreign type, I'm not allowed to `impl Display for ContextError<StringContext>`.

An easy way to use `String` type context errors that may be used in `parse`, so that we can write dynamic error messages that're build during runtime.

The nicest solution would be to provide an `StringContext` context type upstream. That would make it very ergonomic to just use Strings in error messages and I hope it shouldn't be too much of a hassle to implement.

I tried using `const` functions to create static strings, but sadly the Rust language isn't quite there yet, or at least I couldn't find a non-obstruse way to do that.

While writing this I came up with the idea that `once_cell` could be a solution for my problem at hand. But even if that works, this feature would still be a quality of life improvement, as it would allow easier debugging and error message creation without having to come up with a way of creating static strings first.

Another alternative would be to implement a custom Error type, but that feels like overkill to achieve this.

### Additional Context

_No response_

### LINKED ISSUE #778: Create `Context` based on multiple static arrays during runtime

- [x] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [x] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue%20is%3Aclosed%20reason%3Anot-planned) issues

### winnow version

master

### Describe your use case

When parsing large amounts of possible keywords, adding all keyword variants to an error via `.context` quickly becomes bothersome. 

One quickly ends up with 10+ calls of `.context(StrContext::Expected(StrContextValue::StringLiteral("keyword")))`, which then need to be kept up-to date with the arrays that they refer to. This is cumbersome and error prone when removing/adding keywords to any of these lists.

1. A way to add multiple values to a single `Context` instance.
2. A new `MultipleContext` Type that (name to be discussed), which allows this pattern.

For how this might look like, see this code example:

```rust
use winnow::prelude::*;
use winnow::token::one_of;
use winnow::ascii::alpha1;
use winnow::error::{StrContext, StrContextValue};

fn char_parser(input: &mut &str) -> ModalResult<char> {
    let allowed_characters = ['a', 'b', 'c'];
    // Create a simple parser that only allows a few chars with some error context.
    let mut parser = one_of(allowed_characters)
        .context(StrContext::Label("character"));

    // It's now possible to **add** to that context without another `Context` wrapper.
    parser.add_context(StrContext::Expected(StrContextValue::Description(
        "one of the allowed characters",
    )));

    // Dynamically add context info based on runtime variables.
    for character in allowed_characters {
        parser.add_context(StrContext::Expected(StrContextValue::CharLiteral(character)))
    }

    parser.parse_next(input)
}

fn main() {
    assert_eq!(char_parser.parse("a"), Ok('a'));
    assert!(char_parser.parse("d").is_err());
}
```

in #744, I proposed to add a `StringContext` type that allows using `Strings` so that we can concatenate and build the string ourselves, but I believe the other solution is more ideomatic.

### Additional Context

A proposal on how this might look like can be found in https://github.com/winnow-rs/winnow/pull/777

Nukesor:
Heyo :)

While playing around with parsers some more, I came up with an idea on how to make error handling nicer (at least for `std` users). I'm curious what you think about this proposal!

Resolves #778

This MR adds a new `std`-only feature to `Context`, which allows users to dynamically append error context information based on runtime variables.

This was previously not possible as the only way to dynamically append information was to add another `.context()` call, which wrapped the whole parser in a new type, thereby changing the type signature.

This change adds alters the `Context` in a backwards-compatible way by changing the `Context::context` field to a `Vec<C>` and adding a `add_context` function to add context to said vector.

# Downsides

This is obviously a performance downgrade, as this implies that each context parser initialization will do a heap alloc for the Vector.

If that's an issue, what do you think about a completely new type (e.g. `DynamicContext`) and a new function like `dynamic_context` on `Parser` that's hidden behind `std`?

# Example

How does it look in practice:

```rust
use winnow::prelude::*;
use winnow::token::one_of;
use winnow::ascii::alpha1;
use winnow::error::{StrContext, StrContextValue};

fn char_parser(input: &mut &str) -> ModalResult<char> {
    let allowed_characters = ['a', 'b', 'c'];
    // Create a simple parser that only allows a few chars with some error context.
    let mut parser = one_of(allowed_characters)
        .context(StrContext::Label("character"));

    // It's now possible to **add** to that context without another `Context` wrapper.
    parser.add_context(StrContext::Expected(StrContextValue::Description(
        "one of the allowed characters",
    )));

    // Dynamically add context info based on runtime variables.
    for character in allowed_characters {
        parser.add_context(StrContext::Expected(StrContextValue::CharLiteral(character)))
    }

    parser.parse_next(input)
}

fn main() {
    assert_eq!(char_parser.parse("a"), Ok('a'));
    assert!(char_parser.parse("d").is_err());
}
```

### DIFF
--- a/src/combinator/impls.rs
+++ b/src/combinator/impls.rs
@@ -550,6 +550,9 @@ where
 }
 
 /// [`Parser`] implementation for [`Parser::context`]
+///
+/// Contains a single context value.
+#[cfg(not(feature = "std"))]
 pub struct Context<F, I, O, E, C>
 where
     F: Parser<I, O, E>,
@@ -565,6 +568,27 @@ where
     pub(crate) e: core::marker::PhantomData<E>,
 }
 
+#[cfg(not(feature = "std"))]
+impl<F, I, O, E, C> Context<F, I, O, E, C>
+where
+    F: Parser<I, O, E>,
+    I: Stream,
+    E: AddContext<I, C>,
+    E: ParserError<I>,
+    C: Clone + crate::lib::std::fmt::Debug,
+{
+    pub fn new(parser: F, context: C) -> Self {
+        Context {
+            parser,
+            context,
+            i: Default::default(),
+            o: Default::default(),
+            e: Default::default(),
+        }
+    }
+}
+
+#[cfg(not(feature = "std"))]
 impl<F, I, O, E, C> Parser<I, O, E> for Context<F, I, O, E, C>
 where
     F: Parser<I, O, E>,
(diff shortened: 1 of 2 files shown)


epage:
Note: we generally ask that PRs focus on Implementation while Problems and Solutions  be discussed in the issue.  I considered commenting on the Solution and the trade offs in #744 but that seems only tangentially related as written.  I'm assuming #744 was written focused on a problem with a solution to a different problem, rather than the core problem?

PR #801, opened 2025-07-08
Remove unsafe `mem::transmute` calls

joshlf:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,7 +42,7 @@ dependencies = [
  "anstyle",
  "anstyle-parse",
  "anstyle-query",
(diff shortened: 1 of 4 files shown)


joshlf:
The only concern would be future-proofing against changes that affect the layout; with this change, those would be detected and compile failures. But it's a minor concern; definitely a judgment call as to whether it's worth taking a dependency for.

epage:
I copied these from `bstr` and just looked to confirm what they do and apparently the `transmute` can be replaced with a cast, see https://github.com/BurntSushi/bstr/commit/ef95d8848804b365e477edee53b643d7d503db0e.  I'd be open to updating to match `bstr`.

PR #826, opened 2025-09-27
Clarify docs for `Offset` and `Stream`

onkoe:
## Goal

Improve the `Offset` + `Stream` docs.

## Background

I implemented `Stream` for a type yesterday, but it took quite a lot of time to understand what was going on. The existing docs weren't very clear, so it kinda required me to copy what `&[u8]` and `&str` were doing.

After implementing it successfully, I thought I may as well share what I learned! (and spare myself the time re-learning it in the future)

These changes should clarify what an implementation should do... without being too over the top. If I still sounded a little off, or anything is incorrect, please let me know and I'll fix it up! :D

## Changes

- Improve docs on `winnow::stream::Stream`...
- ...and on `winnow::stream::Offset`.
- Fix a couple broken intradoc links.

## Testing

I know it'll run in CI, but I tested my changes like so:

- [x] `typos` 
- [x] `committed` 
- [x] `cargo doc`

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -134,48 +134,92 @@ where
     }
 }
 
-/// Core definition for parser input state
+/// A stream of input data for a [`Parser`](`crate::Parser`).
+///
+/// Types can implement this trait to become custom input types.
 pub trait Stream: Offset<<Self as Stream>::Checkpoint> + crate::lib::std::fmt::Debug {
-    /// The smallest unit being parsed
+    /// The smallest unit being parsed.
     ///
-    /// Example: `u8` for `&[u8]` or `char` for `&str`
+    /// Examples:
+    ///
+    /// - `u8` for `&[u8]`
+    /// - `char` for `&str`
     type Token: crate::lib::std::fmt::Debug;
-    /// Sequence of `Token`s
+
+    /// A sequence of [`Self::Token`]s.
     ///
-    /// Example: `&[u8]` for `LocatingSlice<&[u8]>` or `&str` for `LocatingSlice<&str>`
+    /// Examples:
+    ///
+    /// - `&[u8]` for `LocatingSlice<&[u8]>`
+    /// - `&str` for `LocatingSlice<&str>`
     type Slice: crate::lib::std::fmt::Debug;
 
-    /// Iterate with the offset from the current location
+    /// An iterator containing a tuple of:
+    ///
+    /// - [`usize`]: the offset from the current location.
+    /// - [`Self::Token`]: the token at the `usize` location.
+    ///
+    /// Note that, since [`Iterator`] is a trait, this type can be anything
+    /// that implements that trait. Therefore, an efficient implementation will
+    /// dynamically create this tuple instead of precomputing it.
     type IterOffsets: Iterator<Item = (usize, Self::Token)>;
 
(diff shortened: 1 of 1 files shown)


src/stream/mod.rs
```diff
@@ -134,48 +134,92 @@ where
     }
 }
 
-/// Core definition for parser input state
+/// A stream of input data for a [`Parser`](`crate::Parser`).
+///
+/// Types can implement this trait to become custom input types.
```
epage, line 139: This feels redundant
```suggestion
```

src/stream/mod.rs
```diff
@@ -134,48 +134,92 @@ where
...
 
-    /// Iterate with the offset from the current location
+    /// An iterator containing a tuple of:
+    ///
+    /// - [`usize`]: the offset from the current location.
+    /// - [`Self::Token`]: the token at the `usize` location.
+    ///
+    /// Note that, since [`Iterator`] is a trait, this type can be anything
+    /// that implements that trait. Therefore, an efficient implementation will
+    /// dynamically create this tuple instead of precomputing it.
```
epage, line 164: This feels oddly specific and more about teaching Rust than explaining this functionality

src/stream/mod.rs
```diff
@@ -134,48 +134,92 @@ where
...
     fn iter_offsets(&self) -> Self::IterOffsets;
 
-    /// Returns the offset to the end of the input
+    /// Returns the number of bytes until the stream completes.
+    ///
+    /// That is: the end of the file ("eof").
     fn eof_offset(&self) -> usize;
 
-    /// Split off the next token from the input
+    /// Takes the next token from the stream.
```
epage, line 187: This doesn't express that we are also advancing the stream

src/stream/mod.rs
```diff
@@ -134,48 +134,92 @@ where
...
     ///
-    /// This means "0 tokens" will return `0` offset
+    /// The offset is relative to the current position. This means that an
+    /// input of "0 tokens" will return an offset of `0`.
     fn offset_at(&self, tokens: usize) -> Result<usize, Needed>;
-    /// Split off a slice of tokens from the input
+
+    /// Takes a slice of `offset` tokens from the input.
+    ///
+    /// This method should consume `offset` tokens.
```
epage, line 215: I feel like saying "split off" covers both sentences and ensures its all in the summary

src/stream/mod.rs
```diff
@@ -185,19 +229,21 @@ pub trait Stream: Offset<<Self as Stream>::Checkpoint> + crate::lib::std::fmt::D
...
     /// * Indexes must be within bounds of the original input;
-    /// * Indexes must uphold invariants of the stream, like for `str` they must lie on UTF-8
-    ///   sequence boundaries.
-    ///
+    /// * Indexes must uphold invariants of the stream. For example, with a
+    ///   `str`, they must lie on UTF-8 sequence boundaries.
     fn next_slice(&mut self, offset: usize) -> Self::Slice;
-    /// Split off a slice of tokens from the input
+
+    /// Takes a slice of `offset` tokens from the input.
```
epage, line 239: FYI this is the point  that I'm taking a break.  Please consider my feedback above on the documentation below and we can re-evaluate once its all been updated

PR #846, opened 2025-11-20
docs: Fix some typos

clint-white:
I noticed what look like a couple of typos while reading the docs:
- A sentence in the tutorial that appears to be missing a word.
- Two sentences in `Parser::parse()` that were run together without separator.  From the layout of the comment in the source code, it appears the intent was for the second of those to begin a new paragraph so in addition to the period I added a newline as well.

### DIFF
--- a/src/_tutorial/chapter_7.rs
+++ b/src/_tutorial/chapter_7.rs
@@ -714,7 +714,7 @@
 //!
 //! To add spans to your parsed data for inclusion in semantic errors, see [`Parser::with_span`].
 //!
-//! For richer syntactic with spans,
+//! For richer syntactic errors with spans,
 //! consider separating lexing and parsing and annotating your tokens with [`Parser::with_span`].
 
 #![allow(unused_imports)]


src/parser.rs
```diff
@@ -54,7 +54,8 @@ pub trait Parser<I, O, E> {
     /// [incremental parsing][StreamIsPartial], see instead [`Parser::parse_next`].
     ///
     /// This assumes the [`Parser`] intends to read all of `input` and will return an
-    /// [`eof`][crate::combinator::eof] error if it does not
+    /// [`eof`][crate::combinator::eof] error if it does not.
+    ///
```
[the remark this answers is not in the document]
clint-white: Corrected.

src/parser.rs
```diff
@@ -54,9 +54,8 @@ pub trait Parser<I, O, E> {
     /// [incremental parsing][StreamIsPartial], see instead [`Parser::parse_next`].
     ///
     /// This assumes the [`Parser`] intends to read all of `input` and will return an
-    /// [`eof`][crate::combinator::eof] error if it does not
-    /// To ignore trailing `input`, combine your parser with a [`rest`][crate::token::rest]
-    /// (e.g. `(parser, rest).parse(input)`).
+    /// [`eof`][crate::combinator::eof] error if it does not. To ignore trailing `input`, combine
+    /// your parser with a [`rest`][crate::token::rest] (e.g. `(parser, rest).parse(input)`).
```
epage: If you were just adding a period, why did you reflow this?

As a side note, I tend to use https://sembr.org/ to reduce noise in reviews.

clint-white: Sorry, I did not know about this style of semantic line breaks and that was what was intended here.  Makes sense.  This version now only adds the missing period, otherwise should be exactly the same as before.  Thanks for pointing out the convention.

Issue #847, opened 2025-11-26
Tab characters causes error arrow to be displayed in wrong place


j-g00da:
### Please complete the following tasks

- [x] I have searched the [discussions](https://github.com/winnow-rs/winnow/discussions)
- [x] I have searched the [open](https://github.com/winnow-rs/winnow/issues) and [rejected](https://github.com/winnow-rs/winnow/issues?q=is%3Aissue%20is%3Aclosed%20reason%3Anot-planned) issues

### rust version

rustc 1.91.1 (ed61e7d7e 2025-11-07)

### winnow version

0.7.13; main

### Minimal reproducible code

```rust
use winnow::{
    ModalResult, Parser,
    combinator::{cut_err, eof, terminated},
    error::{StrContext, StrContextValue},
    token::take_while,
};

fn anything_except_at(input: &mut &str) -> ModalResult<String> {
    terminated(cut_err(take_while(0.., |c: char| c != '@')), eof)
        .context(StrContext::Expected(StrContextValue::Description(
            "anything but '@'",
        )))
        .map(|s: &str| s.to_string())
        .parse_next(input)
}

fn main() {
    let input = "\tshould error on the at sign -->@<--";
    let result = anything_except_at.parse(input);

    println!("{}", result.unwrap_err());
}
```

### Steps to reproduce the bug with the above code

Just add winnow to dependencies and `cargo run`.

### Actual Behaviour

```
        should error on the at sign -->@<--
                                ^
expected anything but '@'
```

### Expected Behaviour

```
        should error on the at sign -->@<--
                                       ^
expected anything but '@'
```

### Additional Context

I'm open to work on this.

I think the right way to handle this would be to add tab characters in the 'arrow line' in the same places as in the actual line that the error originated from.

epage:
The error messages are intentionally not fully fleshed out.  If an application is doing high quality error messages, then they should likely be using a package like `annotate-snippets`, `miette`, or `ariadne`.  At that point, anything we do here becomes redundant and slows down compile times.  The errors here are more for quick and dirty reporting.

PR #848, opened 2025-11-26
fix: Correctly align arrow in `ParseError` display when input contains tabs

j-g00da:
Fixes issue with misaligned arrow in `ParseError` when the line from which the error originated contains tab characters.

Closes: #847


j-g00da:
Note: this only fixes the issue with `\t`. Unicode characters treated as being the same width is a different issue.

PR #861, opened 2025-12-31
Fix function documentation to reference postfix and infix correctly

dsemi:
Small documentation fix for some functions in the newly added Pratt parser.


epage:
Thanks!

Issue #794, opened 2025-06-17
Implementing copy for some structures


yyy33:
Currently parser defaults to implementing related structures such as Map, and I think it's possible to implement copy for them, since most parsers exist as free functions and are copied

hgiesel:
@epage had some thoughts [here](https://github.com/winnow-rs/winnow/pull/915#issuecomment-4440654956).

My answers to that are as follows:

> In terms of intentionally providing Copy or Clone, we wouln't be doing that by forking existing parsers. We'd need to change the existing parsers.

Totally agree! I was just testing the waters of what that would look like. The issue is that we'd either break them for users, because the `IgnoredParser`s couldn't return arbitrary types anymore, but rather `()` (Or should the Ignored parsers' output types be put into the `PhantomData` as well?)

My motivation for this issue/PR was that I tried to write my own higher-level parser combinator for parsing pinyin. I try to take the parsers by argument as `pub fn f<'a, T: Clone>(parser_a: impl Parser<&'a str, T, ContextError> + Clone`, ...)`.

But as soon as I do that, I have to wrap the parsers in function, so they can be accepted:

```rs
fn parse_a(input: &mut &str) -> Result<Tone> {
  preceded('a', tone_number).parse_next(input)
}

fn parse_e(input: &mut &str) -> Result<Tone> {
  preceded('a', tone_number).parse_next(input)
}

f(
  parse_a,
  parse_e,
  ...
)
```

What I'd like to write instead would be:

```rs
f(
  preceded('a', tone_number),
  preceded('e', tone_number),
  ...
)
```

Afaik, having a named type as the return type is the only way to accomplish this.

epage:
> The issue is that we'd either break them for users, because the IgnoredParsers couldn't return arbitrary types anymore, but rather () (Or should the Ignored parsers' output types be put into the PhantomData as well?)

We need to track every aspect of every type in `PhantomData` for type inference to work correctly.  See also #222.

> Afaik, having a named type as the return type is the only way to accomplish this.

From https://github.com/winnow-rs/winnow/pull/915#issuecomment-4440654956

>  I've been hesitant to use structs in winnow because `impl Parser` is easier to read and write, making Winnow itself a good example for users in how to do things. Moving away from that should be discussed in the issue first.

[...]

PR #914, opened 2026-05-13
feat: Void preserves clone

hgiesel:
This preserves `Clone` when the callee of `.void()` is `Clone`.

I noticed the codebase doesn't use derive, so I also refrained.

Related: https://github.com/winnow-rs/winnow/issues/794

### DIFF
--- a/src/combinator/impls.rs
+++ b/src/combinator/impls.rs
@@ -396,6 +396,15 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


epage:
Doesn't look like the other parser structs do this and I would rather we be all or nothing, so I'm going to close this as we work things out in #794

PR #915, opened 2026-05-13
feat: add prec combinator

hgiesel:
This is the concept of an alternative to `preceded` that preserves `Clone` on its output by having it's own struct.

The drawback is that it now requires `()` on the ignored parsers' output.

Addresses partially #794.

### DIFF
--- a/src/combinator/mod.rs
+++ b/src/combinator/mod.rs
@@ -188,7 +188,7 @@ pub use self::multi::separated_foldr1;
 pub use self::multi::{
     fill, iterator, repeat, repeat_till, separated, separated_foldl1, ParserIterator, Repeat,
 };
-pub use self::sequence::{delimited, prec, preceded, separated_pair, terminated};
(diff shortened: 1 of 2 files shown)


hgiesel:
Is there any reason why the parsers in `winnow::combinator::sequence` don't already do this, whereas others do?

epage:
Some parsers are structs because
- `trait Parser` can't return `impl Parser` yet
- Functions exist on the struct

In terms of intentionally providing Copy or Clone, we wouln't be doing that by forking existing parsers. We'd need to change the existing parsers. I've been hesitant to use structs in winnow because `impl Parser` is easier to read and write, making Winnow itself a good example for users in how to do things. Moving away from that should be discussed in the issue first.

We would also need to change all existing parsers, not just some, and put it in our contrib docs on writing parsers.

Considering that the focus of discussion should be the Issue, this is incomplete, and this would need a rewrite, I'm going to close this for now.

PR #916, opened 2026-05-13
Remove AsBStr bound from winnow::ascii::float

Uhrendoktor:
The `winnow::ascii::float` methods currently enforce `I: AsBStr`. 

I could not find a feature flag or other reason that would enforce this. 

Removing this would allow for better easier interaction when e.g. parsing single `char`s. Therefore i propose to remove this bound.


epage:
Hard to tell when bounds are no longer used. I wonder if clippy can lint for that. May be worth opening an issue asking for it.

PR #931, opened 2026-07-20
Document format expected by ascii::{dec_int, dec_uint}

wojciech-graj:


epage:
Could you help me understand the point of confusion you had that led to this PR?

wojciech-graj:
I assumed it worked like the FromStr implementations for these types (and we are using FromStr under the hood, after narrowing down which part of the buffer to parse from), which allow leading zeros and a leading + for unsigned integers.

epage:
Thanks! That helps shed light on what needs to be clarified.

Personally, I think helpers like these are only good for prototyping. For production use, people should likely use parsers that map to their BNF easily to make sure they get the right behavior (e.g. should it align with Rust or not?)

wojciech-graj:
> Personally, I think helpers like these are only good for prototyping.

I agree, hence why I don't propose changing these parsers' behavior to match FromStr. But I think their behavior should be documented so users know what to expect, even in the prototyping stage.

epage:
What are your thoughts on us representing the syntax using BNF in a code block?

wojciech-graj:
It's the most precise way to specify the syntax, and it can be used to document any parsers, not just ones where the format is trivial to describe in words, so I'm for it. As a follow-up to this PR, we should consider adding this documentation to more functions.

Even though it's a bit more verbose, I stuck to BNF over EBNF because it's a lot more obvious to readers.

I pushed a commit with this, let me know what you think.

epage:
Thnks! I made some tweaks and merged this in #932.

